PR #52, opened 2021-05-17
Added keep MediaStream while reconnecting Publisher.

R-Delfino95:
Fix multiple event listeners while reconnecting.

### DIFF
--- a/packages/millicast-sdk/src/Publish.js
+++ b/packages/millicast-sdk/src/Publish.js
@@ -124,4 +124,9 @@ export default class Publish extends BaseWebRTC {
     this.setReconnect()
     logger.info('Broadcasting to streamName: ', this.streamName)
   }
+
+  reconnect () {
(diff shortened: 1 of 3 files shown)


packages/millicast-sdk/src/Publish.js
```diff
@@ -124,4 +124,9 @@ export default class Publish extends BaseWebRTC {
     this.setReconnect()
     logger.info('Broadcasting to streamName: ', this.streamName)
   }
+
+  reconnect () {
+    this.options.mediaStream = (this.webRTCPeer?.peer) ? this.webRTCPeer.peer.getSenders().map((sender) => sender.track) : this.options.mediaStream
```
emilsas, line 129: It would be better to mainting RTCPeerConnection access inside our PeerConnection class. Consider a new method in PeerConnection called `getActiveTracks()`

emilsas (approved):

Issue #59, opened 2021-05-22
Audio not working when published from this SDK using h264 and Chrome browser.


jlcastrillon91:
When trying to broadcast from Chrome browser using the 'h264' codec, I can't hear the audio from millicast hosted viewer (the video can be seen fine).

I am using the following code.
```
async broadcastStream(stream) {
    let streamName = `${this.options.roomName}`;
    //Authenticate publisher and get connection point.
    const tokenGenerator = () => Director.getPublisher(this.options.millicastPublishToken, streamName)
    //Create a new instance
    const millicastPublish = new Publish(streamName, tokenGenerator)
    //Publishing Options
    const broadcastOptions = {
        mediaStream: stream,
        simulcast: options.useSimulcast,
        disableAudio: false,
        diableVideo: false,
        bandwidth: 0,
        codec: 'h264'
    }
    //Start broadcast
    return millicastPublish.connect(broadcastOptions);
}
```

When using 'vp8' codec the audio works fine.

Also if using the publiser demo from this repository I have the same issue if I change the vp8 codec to h264 directly from code
![image](https://user-images.githubusercontent.com/39222433/119233202-dcda9b00-baf5-11eb-908f-81e0931b212e.png)

This is the setup I am using:
**Operating System:** Windows 10, 64 bits
**Browser:** Chrome, version 90.0.4430.212 (Official build) (64 bits)
**Package version:**  "@millicast/sdk": "^0.1.3"

emilsas:
Hi @jlcastrillon91, let us check this issue by ourselves.

murillo128:
The problem is when mungling the SDP. This is the offer we use for SLD:

```
sdp: v=0
o=- 3658879760281266610 2 IN IP4 127.0.0.1
s=-
t=0 0
a=group:BUNDLE 0 1
a=extmap-allow-mixed
a=msid-semantic: WMS YaC1dzxHxTgF9czkqspjbjF5wjfxfAUDugVG
m=audio 9 UDP/TLS/RTP/SAVPF 111 103 104 9 0 8 106 105 13 110 112 113 126 127
a=rtpmap:127 multiopus/48000/6
a=fmtp:127 channel_mapping=0,4,1,2,3,5;coupled_streams=2;minptime=10;num_streams=4;useinbandfec=1
c=IN IP4 0.0.0.0
a=rtcp:9 IN IP4 0.0.0.0
a=ice-ufrag:hzg1
[...]

PR #60, opened 2021-05-25
Fix: Audio track not working with H.264

emilsas:
Due to #59 issue, enabling multiopus support with H.264 video codec causes that audio stream does not work and receives 0 bytes. 
This behaviour occurs with a 5.1 stream and with most common streams with max 2 channels. We are currently investigating why this happens, but now we decided to disable the multi-channel feature only with H.264.

## Update
The problem was that we are using the same payload type id of H.264 with multiopus. This was resolved calculating all available payload type ids and choosing one of them.


R-Delfino95 (approved):

murillo128:
lgtm

PR #66, opened 2021-06-01
Feature/refactor peer connection stats

R-Delfino95:


packages/millicast-sdk/src/PeerConnectionStats.js
```diff
@@ -8,23 +8,38 @@ const logger = Logger.get('PeerConnectionStats')
  * @property {RTCStatsReport} raw - All RTCPeerConnection stats without parsing.
  * @property {TrackReport} audio - Parsed audio information.
  * @property {TrackReport} video - Parsed video information.
+ * @property {Number} availableOutgoingBitrate - Available outgoing bitrate.
```
emilsas, line 11:
```suggestion
 * @property {Number} availableOutgoingBitrate - The available outbound capacity of the network connection. The higher the value, the more bandwidth you can assume is available for outgoing data. The value is reported in bits per second.
```

packages/millicast-sdk/src/PeerConnectionStats.js
```diff
@@ -8,23 +8,38 @@ const logger = Logger.get('PeerConnectionStats')
  * @property {RTCStatsReport} raw - All RTCPeerConnection stats without parsing.
```
emilsas, line 8:
```suggestion
 * @property {RTCStatsReport} raw - All RTCPeerConnection stats without parsing. Reference https://developer.mozilla.org/en-US/docs/Web/API/RTCStatsReport.
```

emilsas (changes requested):
I will add in jsdocs of reports, a description of each field based on MDN documentation.

emilsas (approved):

PR #67, opened 2021-06-02
Add multiopus codec at the end of the rtpmaps for audio

murillo128:


emilsas:
I tested your fix of multiopus in my Mac. Publishing a stereo/mono audio track works fine. But multichannel tracks stopped working in Chrome, it detects opus.
So the only way to enable multiopus is detecting if the AudioTrack attached has more than 2 channels.
I closed this PR because we need to make another fix.

PR #68, opened 2021-06-03
Fix: EventSubscriber disconnection.

R-Delfino95:
- Added conditional for close Event Subscriber webSocket if exists.
- Added scenario and test case.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,3 @@ coverage
 dist
 docs
 .env
-env.js
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ coverage
 dist
 docs
 .env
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -3,4 +3,5 @@ node_modules
 coverage
 dist
 docs
-.env
\ No newline at end of file
+.env
+env.js
```
emilsas, line 7: This could break env.js of publisher and viewer demo packages

emilsas (changes requested):

.gitignore
R-Delfino95: Removed.

emilsas (approved):

Issue #65, opened 2021-06-01
Audio not working in Safari when the publisher is using h264 codec.


CastriOnlive:
**Description of the bug**
URGENT: The audio is not working when publishing with h264 codec and viewing the stream with Safari (iOS and Mac).

**Steps to reproduce**
1. Install latest version of publisher demo [](https://github.com/millicast/millicast-sdk/tree/main/packages/millicast-publisher-demo) with npm.
2. Change src/publisher.js to h264 codec (see screenshot with code).
3. Set .env with the necessary environment tokens and a stream name.
4. Run npm start
5. Start broadcast
6. Open the millicast hosted viewer and unmute the audio in the controls [](http://viewer.millicast.com/v2?streamId={IPlacedMyAccountIdHere}/{IPlacedTheStreamNameHere})

**Expected behavior**
When unmuting the audio in the hosted viewer controls it should be possible to listen to the audio in the broadcast.

**Screenshots**
![image](https://user-images.githubusercontent.com/85135004/120332373-aef71280-c2bc-11eb-9e08-2312e11b8309.png)
![image](https://user-images.githubusercontent.com/85135004/120339401-1748f280-c2c3-11eb-924f-9fd85df8891a.png)
![image](https://user-images.githubusercontent.com/85135004/120340764-4e6bd380-c2c4-11eb-91b2-4079ea656b91.png)

**Environment:**
_Publisher_
- OS: Windows 10, Also tested with Ubuntu 20.04
- Browser: Chrome 
- Version 90.0.4430.212 (Official Build) (64 bits)
_Viewer_
- OS: iOS, Mac
- Browser: Safari 
- Version 14.0

**Additional context**
Also tested using the SDK directly without the publisher demo.
```
  //Authenticate publisher and get connection point.
  const tokenGenerator = () => Director.getPublisher(publishToken, streamName)
  //Create a new instance
  const millicastPublish = new Publish(streamName, tokenGenerator)
  //Publishing Options
  const broadcastOptions = {
      mediaStream: stream,
      simulcast: true,
      disableAudio: false,
      diableVideo: false,
      bandwidth: 0,
      codec: 'h264'
  }
  //Start broadcast
  return millicastPublish.connect(broadcastOptions);
```

Also tried using the viewer of the SDK and got the same result
```
async receiveStream(streamName, accountId) {
    return new Promise(async (resolve, reject) => {
        //Define callback for generate new token
[...]

PR #69, opened 2021-06-04
Fix: Set multiopus if current MediaStream has more than 2 channels.

R-Delfino95:
Due to #65, we set multiopus in the SDP if MediaStream is multichannel or is a Viewer (no MediaStream).
Added scenarios and tests.
Renamed env-sample variables to test-enviroment-samples.

### DIFF
--- a/packages/millicast-sdk/src/utils/SdpParser.js
+++ b/packages/millicast-sdk/src/utils/SdpParser.js
@@ -172,12 +172,13 @@ export default class SdpParser {
    *
    * **Only available in Google Chrome.**
    * @param {String} sdp - Current SDP.
+   * @param {MediaStream} mediaStream - MediaStream offered in the stream.
    * @returns {String} SDP parsed with multiopus support.
-   * @example SdpParser.setMultiopus(sdp)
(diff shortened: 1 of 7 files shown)


packages/millicast-sdk/src/utils/SdpParser.js
```diff
@@ -221,3 +222,12 @@ export default class SdpParser {
...
   }
 }
+
+/**
+ * Checks if mediaStream has more than 2 audio channels.
+ * @param {MediaStream} mediaStream - MediaStream to verify.
+ * @returns {Boolean} returns true if MediaStream has more than 2 channels.
+ */
+const hasAudioMultichannel = (mediaStream) => {
+  return !!mediaStream.getAudioTracks().find(value => value.getSettings().channelCount > 2)
```
emilsas, line 232:
```suggestion
  return mediaStream.getAudioTracks().some(value => value.getSettings().channelCount > 2)
```

emilsas (changes requested):

emilsas (approved):

PR #78, opened 2021-06-25
Feature: Record option refactor

R-Delfino95:
- If record option is not provided, use default token configuration. 
- Added behaviour in documentation.


murillo128:
lgtm

PR #99, opened 2021-09-16
Release v0.1.12

R-Delfino95:
# Changelog

### Added
* View and Publish modules
  * Add new `absCaptureTime` option in `.connect()` methods. This support abs-capture-time by allowing to mungle the SDP to add the extamp for the extension before the SLD is called.


nico-ramos (approved):

santi:
Great job on patching the lookbehind/lookahead regex into a more compatible regex for replacing domain name  (was leading to crashes in Safari). Keep up the great work 💯

PR #120, opened 2021-11-08
Release v0.1.16

R-Delfino95:


packages/millicast-sdk/src/Publish.js
```diff
@@ -110,7 +111,7 @@ export default class Publish extends BaseWebRTC {
       logger.error('Error while broadcasting. Publisher data required')
       throw new Error('Publisher data required')
     }
-    const recordingAvailable = jwtDecode(publisherData.jwt).millicast.record
+    const recordingAvailable = jwtDecode(publisherData.jwt)[atob('bWlsbGljYXN0')].record
```
santi, line 114: Is there any reason to decode the base64-encoded string 'millicast' and use it as a key on the decoded JWT? To me this looks like unnecessary complexity for a single value lookup

santi: @nicolaslevy

Issue #122, opened 2021-11-22
Dep `rollup-plugin-inject-process-env` in v0.1.16+ makes SDK unusable in standard Vue / webpack setup


santi:
**Version affected**
v0.1.16

**Describe the bug**
The newly introduced plugin `rollup-plugin-inject-process-env` (used to inject env variables during build) is incompatible with webpack production builds using common config setups. 

This is due to `rollup-plugin-inject-process-env` injecting code at the top of output modules, with the following line: 
`process.env = Object.assign({}, process.env);`

This line is further processed by webpack, where it by default replaces references to `process.env` with an object containing the actual env variables. The line of code from rollup is thus converted into: `{NODE_ENV:"production",MORE_ENV_VARIABLES:"example"} = Object.assign({},{NODE_ENV:"production",MORE_ENV_VARIABLES:"example"})`

Which is not valid JS (you are basically trying to destructure the result from `Object.assign` into an object with values). 

Note: This is NOT a problem during development config with webpack, only when building for production / with minification enabled.

**Proposed fix**
Use another library or method to use env variables in code. This library is also reported to have [problems with standard setups of Create React app](https://github.com/badcafe/rollup-plugin-inject-process-env/issues/13): 

**To Reproduce**
Example repo: 
[Rollup env transpilation error](https://github.com/santi/rollup-env-transpilation-error)

1. Create a basic Vue 3 project with Vue CLI: `npx @vue/cli create example-project`
2. Install Millicast SDK: `cd example-project && yarn add @millicast/sdk`
3. Add an import from Millicast SDK and use it somehow: `import {View} from "@millicast/sdk"; console.log(View);`
4. Try to build project: `yarn build`
5. Expected error: 
```
js/chunk-vendors.5cc891d8.js from Terser
Invalid assignment [./node_modules/@millicast/sdk/dist/millicast.umd.js:1,647][js/chunk-vendors.5cc891d8.js:13296,572]
```

**Expected behavior**
Minified build including the Millicast SDK library

**Screenshots**

**Environment (please complete the following information):**
 - OS: macOS 12.0.1
 - Browser: Node
 - Version 16/17 (unrelated)

**Additional context**

R-Delfino95:
Thanks for report it. Yes, we should change how we inject the environment variables, we are searching for an alternative.

For now you can use the [v0.1.15](https://github.com/millicast/millicast-sdk/releases/tag/v0.1.15) that has the same features as the new one.

santi:
[...]

PR #123, opened 2021-11-24
Use @rollup/plugin-replace to support env variables in browser lib build

santi:
Solves https://github.com/millicast/millicast-sdk/issues/122 (see comments for proposed solution which this PR implements)

This PR changes the build output to keep the `process.env.X` entries in the `cjs` and `esm` outputs, while replacing them with undefined in the `umd` output (which is the output used by your CDN). Bundlers will then be able to replace instances of `process.env` while bundling the consuming app, and using the UMD build from the CDN directly will fallback to the default values (by using `||`).


cantuccis (approved):

murillo128:
we need a contributor agreement signed before being able to merge the PR, we will rollback the changes and merge them back as soon as we can get the CA signed.

You are the first contributor outside millicast, so sorry for the inconvenience 🙂

R-Delfino95:
@santi apologies for the inconvenience, could you make a new PR with these changes?
We are currently working in adding a CA assistant so you can accept the contributor agreements and we can merge it.

santi:
@R-Delfino95 Sure, no worries :blush:

PR #137, opened 2022-02-02
Release v0.1.21

nico-ramos:
# Changelog

### Changed
- Removed mediaId null check in `View.project` and added the mediaId doc: _If no mediaId is defined, the first track from the main media stream with the same media type as the input source track will be used_


murillo128 (approved):
lgtm

PR #161, opened 2022-05-11
Feature/remove streamevents

fcancela:
- Removed `StreamEvents` class.
- Removed `EventSubscriber` class.
- Added `viewercount` to the list of broadcast events in the publisher example app.


packages/millicast-publisher-demo/src/publisher.js
```diff
@@ -71,12 +71,18 @@ document.addEventListener("DOMContentLoaded", async (event) => {
...
   //////
 
-  // Add UserCount event listener
-  const events = await StreamEvents.init()
-  events.onUserCount(accountId, streamName, ({count}) => {
-    userCount.innerHTML = count
-  })
-  
+  function setUserCount() {
+    console.log('enter')
```
sebastianpiq, line 75:
```suggestion
```

sebastianpiq (changes requested):

nico-ramos (approved):

PR #170, opened 2022-06-22
Enable the SDK to run in react-native

Yousif-CS:
This is a patch to enable the SDK to run in react-native environments, where a custom JavaScriptCore
engine is used with no name. By checking whether the browser name is defined or not, we can determine whether we are running inside the custom JavaScriptCore Engine or in a normal browser without needing to add `react-native` as a dependency.


murillo128:
@sebastianpiq @nico-ramos @R-Delfino95 ptal

PR #172, opened 2022-07-05
Set new signaling and pc after ws is connected and not after the subscription is sent

murillo128:


murillo128:
maybe the signaling should be set after the connect, but the webrtcPeer, after the subscribe, wdyt?

fcancela (approved):

PR #228, opened 2023-07-27
Feature/switching sources and quality selector features added to Multiview Demo

NicolasInzua:


packages/millicast-multiview-demo/multi.js
```diff
@@ -20,68 +24,196 @@ viewer.on('broadcastEvent', (event) => {
...
+
+const removeSourceOption = (sourceId) => {
+  const sourceToRemove = sourcesDropDown.querySelector(`option[value="${sourceId}"]`)
+  if (sourceToRemove) {
+    sourcesDropDown.removeChild(sourceToRemove)
+  }
+}
+
+sourcesDropDown.addEventListener('change', () => {
+  layersDropDown.innerHTML = ''
```
fcancela, line 206: It would be nice to have if we could store all the layer info. This causes the layers dropdown to stay empty until a new layers event arrives.

fcancela (approved):
Just a reminder to follow the structure of the different packages we have in the repo. 

It can be added incrementally later on. So I don't think it's necessary to block the PR. But make sure to add the missing files afterward. It should include package.json, rollup.config.js, prepare.js, env.js, etc...

packages/millicast-multiview-demo/rollup.config.js
```diff
@@ -0,0 +1,58 @@
...
+
+const environment = getEnvironment()
+
+let watchPlugins = []
+if (process.env.ROLLUP_WATCH) {
+  watchPlugins = [
+    serve({
+      open: true,
+      contentBase: 'dist',
+      port: 10002
```
fcancela, line 19: Change the port to one that's not being used by other apps.

fcancela (commented):
Change the port and it's good to go!

PR #233, opened 2023-08-23
Add H.265 support if the browser supports H.265

dubeyShivank:
Previously the SDK was able to broadcast with H.265 but was not able to publish it when queried with `PeerConnection.getCapabilities('video')`. 

Now the SDK will return `H.265` if the browser supports it. 

Tested on Chrome, Edge, Firefox and Safari on MacOS - h.265 only showed up on safari when the experimental flag was enaabled.


Santiago-Souto (approved):
LGTM

PR #277, opened 2023-10-17
Updated readme files for sample applications

mhelf-dolby:


Santiago-Souto:
@mhelf-dolby for future Pull Requests, please make sure to branch out from `develop` branch and create the PR to there. I this case there is no difference between both, but in other cases that should imply that the PR to main has to be closed and a new one needs to be created to develop.

Santiago-Souto (approved):
LGTM

PR #296, opened 2023-12-11
Fix addRemoteTrack promise not resolving on Safari

R-Delfino95:
- Refactor the implementation of `addRemoteTrack` to have retries on getting the `mid` transceiver and fail if they reach 10 retries.

### DIFF
--- a/packages/millicast-sdk/src/PeerConnection.js
+++ b/packages/millicast-sdk/src/PeerConnection.js
@@ -160,22 +160,17 @@ export default class PeerConnection extends EventEmitter {
    * @return {Promise<RTCRtpTransceiver>} Promise that will be resolved when the RTCRtpTransceiver is assigned an mid value.
    */
   async addRemoteTrack (media, streams) {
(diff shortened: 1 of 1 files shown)


Santiago-Souto (approved):
LGTM

packages/millicast-sdk/src/PeerConnection.js
```diff
@@ -513,6 +501,26 @@ const addReceiveTransceivers = (peer, options) => {
...
 
+const getTransceiverWithMid = async (transceiver, streams, retries = 0) => {
+  return new Promise((resolve, reject) => {
+    if (transceiver.mid) {
+      for (const stream of streams) {
+        stream.addTrack(transceiver.receiver.track)
+      }
+      resolve(transceiver)
+    } else if (retries >= 10) {
+      reject(new Error('Error, maximum number of retries has been reached'))
```
vincentsong, line 512: should add `return` after`reject`

PR #300, opened 2023-12-29
Rollback iCloud Private Relay

Santiago-Souto:
The fix was made for all Safari users, which impoverishes the user experience by being redirected to a TURN server that give worse network connection. To solve the issue, a fix has been made in the server to avoid getting the original bug.


R-Delfino95 (approved):

cdemi:
Hi @Santiago-Souto, this breaks playback on iOS devices with iCloud Private Relay enabled for us

https://github.com/millicast/millicast-sdk/assets/8025435/3e6e4852-d9dd-4667-a317-d148087de7a8

PR #302, opened 2024-01-11
Add description of DTX purpose

Santiago-Souto:
Also, fix typos & update docs


vincentsong:
Should we also review the translation json files? If not, Could we ignore the generated translation folder  `docs-translations`?

vincentsong (commented):
Comments are looks good to me.  But should we also review the translation json files? If not, Could we ignore the generated translation folder `docs-translations`?

R-Delfino95:
> Comments are looks good to me. But should we also review the translation json files? If not, Could we ignore the generated translation folder `docs-translations`?

In the deployment the workflow generates the translation folder. I think It's not necessary to build it each time we modify the docs.
We could add the doc-translation to .gitignore and remove the folder.

Santiago-Souto:
I left #305 open to remove this unnecessary directory. Thanks @vincentsong for pointing it out! Once that is merged we will pull develop and remove does changes.

R-Delfino95 (approved):

PR #304, opened 2024-01-11
[Doc]: Default values and example function name in jsdoc of publish connect function

vincentsong:
1. added default values for connect options
2. renamed to getYourPublisherConnection function in example code


Santiago-Souto (approved):
LGTM

PR #306, opened 2024-01-17
First approach of a diagnostics report Logger function

Santiago-Souto:
- As an exported function, Logger.diagnose() returns an object with relevant information for debugging purposes.
- As an exported variable, Logger.VERSION returns the current SDK version.

### DIFF
--- a/packages/millicast-sdk/src/Director.js
+++ b/packages/millicast-sdk/src/Director.js
@@ -1,5 +1,6 @@
 import Logger from './Logger'
 import config from './config'
+import Diagnostics from './utils/Diagnostics'
 import FetchError from './utils/FetchError'
 
 const logger = Logger.get('Director')
@@ -176,6 +177,7 @@ const Director = {
 
   getSubscriber: async (options, streamAccountId = null, subscriberToken = null) => {
     const optionsParsed = getSubscriberOptions(options, streamAccountId, subscriberToken)
+    Diagnostics.setAccountId(optionsParsed.streamAccountId)
     logger.info(`Getting subscriber connection data for stream name: ${optionsParsed.streamName} and account id: ${optionsParsed.streamAccountId}`)
 
     const payload = { streamAccountId: optionsParsed.streamAccountId, streamName: optionsParsed.streamName }
--- a/packages/millicast-sdk/src/Logger.js
+++ b/packages/millicast-sdk/src/Logger.js
@@ -1,5 +1,6 @@
 import jsLogger from 'js-logger'
 import { version } from '../package.json'
+import Diagnostics from './utils/Diagnostics'
 
 /**
  * @module Logger
@@ -233,6 +234,22 @@ const Logger = {
    * Logger.setHandler(myHandler, Logger.INFO)
    */
   setHandler: (handler, level) => { customHandlers.push({ handler, level }) },
+  /**
+   * @function
+   * @name diagnose
+   * @description Returns an object with diagnostics about the state of the connection and environment.
+   * @param {Number} [statsCount = 5]  - Amount of stats objects to be saved.
+   * @returns {Object} Relevant information about the current state, such us userAgent, SDK version, besides others.
+   * @example
+   * // Log and get a diagnose object with the last 3 stats reports
+   * const diagnostics = await Logger.diagnose(3)
+   */
+  diagnose: Diagnostics.get,
+  /**
+   * @var
+   * @name VERSION
+   * @description Returns the current SDK version.
+   */
   VERSION: version
 }
 
--- a/packages/millicast-sdk/src/Signaling.js
+++ b/packages/millicast-sdk/src/Signaling.js
@@ -4,6 +4,7 @@ import Logger from './Logger'
 import SdpParser from './utils/SdpParser'
 import { VideoCodec } from './utils/Codecs'
 import PeerConnection from './PeerConnection'
+import Diagnostics from './utils/Diagnostics'
 
 const logger = Logger.get('Signaling')
 
@@ -212,6 +213,10 @@ export default class Signaling extends EventEmitter {
       logger.debug('Command result: ', result)
       this.serverId = result.subscriberId
       this.clusterId = result.clusterId
+
+      // Save for diagnostics
+      Diagnostics.setStreamNameAndSubscriberId(this.streamName, this.serverId)
+      Diagnostics.setStreamViewId(result.streamViewId)
       return result.sdp
     } catch (e) {
       logger.error('Error sending view command, error: ', e)
@@ -289,6 +294,10 @@ export default class Signaling extends EventEmitter {
       logger.debug('Command result: ', result)
       this.serverId = result.publisherId
       this.clusterId = result.clusterId
+
+      // Save for diagnostics
+      Diagnostics.setStreamNameAndSubscriberId(this.streamName, this.serverId)
+      Diagnostics.setFeedId(result.feedId)
       return result.sdp
     } catch (e) {
       logger.error('Error sending publish command, error: ', e)
--- a/packages/millicast-sdk/src/utils/Diagnostics.js
+++ b/packages/millicast-sdk/src/utils/Diagnostics.js
@@ -1,0 +1,60 @@
+import { version } from '../../package.json'
+
+const MAX_STATS_HISTORY_SIZE = 5
+const diagnostics = {
+  version: '',
+  accountId: '',
+  streamName: '',
+  subscriberId: '',
+  timestamp: '',
+  userAgent: '',
+  connection: '',
+  stats: []
+}
+
+const setAccountId = (accountId) => { diagnostics.accountId = accountId }
+
+const setStreamNameAndSubscriberId = (streamName, subscriberId) => {
+  diagnostics.streamName = streamName
+  diagnostics.subscriberId = subscriberId
+}
+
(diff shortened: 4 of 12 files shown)


packages/millicast-sdk/src/Director.js
```diff
@@ -169,6 +169,7 @@ export default class Director {
 
   static async getSubscriber (options, streamAccountId = null, subscriberToken = null) {
     const optionsParsed = getSubscriberOptions(options, streamAccountId, subscriberToken)
+    saveAccountId(optionsParsed.streamAccountId)
```
vincentsong, line 172: Solution1:

Advice to make 'saveAccountId' function as a method of Logger. So we could invoke it like this:
```suggestion
    Logger.saveAccountId(optionsParsed.streamAccountId)
```

Or in oder to avoid coupling/pollute these with Logger, we could use another class/utility to manage the data . 
```
    const Diagnostics = {
       ........
}
```
We only use Logger to print this Object.


-----------------------------

Solutions2:

A better design using interface.  Director own a property `dataListener` which confirms to an interface like this below
```
  interface DataListener {
    updateAccountId(accountId: string)
   <other functions>
}
```
Then make Logger or Diagnostics implement this interface (like solutions1) and we could register it:
```
   Director.registerDataListener(Logger/Diagnostics)
```
Finally, Director can invoke these callbacks of data listener
```
    // inside of Director methods
    this.dataListener?.updateAccountId(accountId)
```
Note: PeerConnectionStats and Signaling need to have their own DataListener interface as well.

Santiago-Souto: I understand this suggestion is meant to look for a clearer way to call these save methods. It is easier to understand that calling `Diagnostics.saveAccountId(accountId)` is meant to save that value for diagnostics purposes.
If that is the case, I think the way to go would be to create a new module named `Diagnostics` (Solution 1), and call those methods from there.
Is that OK?

vincentsong: Yes, sounds good.

Santiago-Souto: @vincentsong please check the latest changes and lets discuss if that is what you meant.

packages/millicast-sdk/src/Logger.js
```diff
@@ -211,7 +213,12 @@ const Logger = {
    *
    * Logger.setHandler(myHandler, Logger.INFO)
    */
-  setHandler: (handler, level) => { customHandlers.push({ handler, level }) }
+  setHandler: (handler, level) => { customHandlers.push({ handler, level }) },
+  /**
+   * Prints out an object with diagnostics about the state of the connection and environment.
+   */
+  diagnose: Diagnostics.get,
```
vincentsong, line 220: Please add @example in JSDoc to demonstrate how to use this new function

packages/millicast-sdk/tests/features/step-definitions/LoggerDiagnose.steps.js
```diff
@@ -0,0 +1,151 @@
...
+const feature = loadFeature('../LoggerDiagnose.feature', { loadRelativePath: true, errors: true })
+
+jest.mock('../../../src/Signaling')
+
+const expectedObject = {
+  accountId: expect.any(String),
+  streamName: expect.any(String),
+  subscriberId: expect.any(String),
+  connection: expect.any(String),
+  version: expect.any(String),
```
vincentsong, line 19: maybe we could use fixed/constant version and streamName for better verification.

Santiago-Souto: streamName cannot be fixed as it does not return the one passed, because the method that saves it is mocked

packages/millicast-sdk/src/Signaling.js
```diff
@@ -78,7 +78,7 @@ export default class Signaling extends EventEmitter {
    */
   async connect () {
     logger.info('Connecting to Signaling Server')
-    if (this.transactionManager && this.webSocket?.readyState === WebSocket.OPEN) {
+    if (this.transactionManager != null && this.webSocket?.readyState === WebSocket.OPEN) {
```
vincentsong, line 81: Is it enough only check `null` without checking `undefined`?  
```suggestion
    if (this.transactionManager && this.webSocket?.readyState === WebSocket.OPEN) {
```

packages/millicast-sdk/src/utils/Diagnostics.js
```diff
@@ -28,21 +35,39 @@ const setStats = (stats) => {
   diagnostics.stats.push(stats)
 }
 
-const getDiagnostics = (statsCount = 5) => {
+const initStats = async (statsCount) => {
+  try {
+    webRTCPeer.initStats()
+    await new Promise(resolve => setTimeout(resolve, statsCount * 1010))
```
vincentsong, line 41: I advice we'd better to listen to the event of stats instead of using a timer here

packages/millicast-sdk/src/utils/Diagnostics.js
```diff
@@ -28,21 +35,39 @@ const setStats = (stats) => {
...
+
+const getDiagnostics = async (statsCount = STATS_HISTORY_SIZE) => {
   diagnostics.version = version
   diagnostics.timestamp = Date.now()
   diagnostics.userAgent = window?.navigator?.userAgent || 'No user agent available'
 
-  if (Number.isInteger(statsCount) && statsCount >= 0 && statsCount < STATS_HISTORY_SIZE) {
-    diagnostics.stats = diagnostics.stats.slice(-statsCount)
+  if (!(diagnostics.stats.length > 0)) {
+    console.log('Getting stats for debugging purposes.')
```
vincentsong, line 54: I advice to use Logger instead of console.log here

packages/millicast-sdk/src/utils/Diagnostics.js
```diff
@@ -28,21 +35,39 @@ const setStats = (stats) => {
...
+}
+
+const getDiagnostics = async (statsCount = STATS_HISTORY_SIZE) => {
   diagnostics.version = version
   diagnostics.timestamp = Date.now()
   diagnostics.userAgent = window?.navigator?.userAgent || 'No user agent available'
 
-  if (Number.isInteger(statsCount) && statsCount >= 0 && statsCount < STATS_HISTORY_SIZE) {
-    diagnostics.stats = diagnostics.stats.slice(-statsCount)
+  if (!(diagnostics.stats.length > 0)) {
```
vincentsong, line 53:
```suggestion
  if (diagnostics.stats.length) {
```
then switch the blocks below is much better

packages/millicast-sdk/src/Logger.js
Santiago-Souto: Please, resolve if the changes are good

packages/millicast-sdk/src/utils/Diagnostics.js
vincentsong: After looked further, I have advices below:
1. this function named `initStats` but it stops Stats after a while by the timer, this will confuse the user, we could provide a `stopStats` function for user
2. I see we have max stats number protection when we record the stats in the callback, so we don't need the count down timer here and we will always get the latest stats and remove the deprecated ones.
3. will the stopStats conflict with user's initStats, I mean user started the stats but have not stopped it yet?

Santiago-Souto: 1. I agree the function does not give full visibility, maybe `getStats` would be better. Anyway, this is not an exposed function to the user, so the user should not be able to use it.
2. and 3. This function is only called when there are no stats available, meaning that the `initStats` was never called, so the timer will be based on how many stats are asked for. Something I will add is to check that the count passed is not bigger than the maximum.

vincentsong: Regarding question 3,  will our `stopStats` lead to unexpected behaviour when user want to keep receiving the stats?

And since we only save the latest N stats records, I don't think we need the timer and call `stopStats`.  I am thinking we should have a global switch for diagnostics, when it's on, we should keep stats open even after user called `stopStats`. But when diagnostics off, it will depends on user's call to turn on/off stats.

vincentsong (commented):
Just updated the conversation https://github.com/millicast/millicast-sdk/pull/306#discussion_r1473610544

And let's talk tomorrow.

Santiago-Souto:
@vincentsong just made the changes we agreed yesterday. Please have a look.

packages/millicast-sdk/src/utils/Diagnostics.js
```diff
@@ -0,0 +1,60 @@
...
+const setConnectionState = (connectionState) => { diagnostics.connection = connectionState }
+
+const setStats = (stats) => {
+  if (diagnostics.stats.length === MAX_STATS_HISTORY_SIZE) {
+    diagnostics.stats.shift()
+  }
+  diagnostics.stats.push(stats)
+}
+
+const getDiagnostics = async (statsCount = MAX_STATS_HISTORY_SIZE) => {
```
vincentsong, line 35: Do we need async function here?
```suggestion
const getDiagnostics = (statsCount = MAX_STATS_HISTORY_SIZE) => {
```

vincentsong (dismissed):

vincentsong (commented):
We also need to update the index.d.ts for the Diagnostics, isn't it?

vincentsong (approved):

PR #307, opened 2024-01-18
Document: remove namespaces and add diagram

vincentsong:
- Replaced namespaces with modules in JSDoc for better navigation
- Added UML class diagram using mermaid syntax


Santiago-Souto (approved):
Looks great! 🚀

PR #310, opened 2024-01-25
Added schema validation for publish connect options

vincentsong:


Santiago-Souto:
@vincentsong Something to take into account is that not all options have to be passed to the connect function. As so, we should be certain that `joi` package when validating does not throw an error when a value is not passed as an option. We should also look for the ones that are mandatory, if any, and see how to handle that as well.

vincentsong:
> @vincentsong Something to take into account is that not all options have to be passed to the connect function. As so, we should be certain that `joi` package when validating does not throw an error when a value is not passed as an option. We should also look for the ones that are mandatory, if any, and see how to handle that as well.

According the doc, `Values (or keys in case of objects) are optional by default.` which means if we didn't pass the parameter in the schema to the function as an option, the error will be undefined, so it's fine. I think we don't have the mandatory option in our `connect` function, isn't it?

Santiago-Souto (changes requested):

Santiago-Souto (approved):
LGTM

PR #311, opened 2024-01-25
Feature / Functional tests E2E

Santiago-Souto:


vincentsong:
I am confused by the folder structure in tests, could we make it more clear by using `unit`, `e2e`?

.github/workflows/check-tests.yml
```diff
@@ -56,8 +56,16 @@ jobs:
             */*/node_modules
           key: ${{ runner.os }}-${{ hashFiles('**/package-lock.json') }}
 
-      - name: Test 🧪
+      - name: Prepare Tests
         working-directory: './packages/millicast-sdk'
         run: |
-          npm install
-          npm test
+          echo -n ${{secrets.MILLICAST_TEST_ACCOUNT}} | base64 -d > ./tests/test-env.js
```
vincentsong, line 62: We don't need use any .env file in Github Workflow.  Instead, we could use env variables in job, such as the `NODE_AUTH_TOKEN` in `new-release` workflow.
Check this: [Defining environment variables for a single workflow](https://docs.github.com/en/actions/learn-github-actions/variables#defining-environment-variables-for-a-single-workflow)

.github/workflows/check-tests.yml
```diff
@@ -56,8 +56,18 @@ jobs:
...
+
+      - name: Unit Testing 🧪
+        working-directory: './packages/millicast-sdk'
+        run: npm run test-unit
+
+      - name: End-2-End Testing
+        working-directory: './packages/millicast-sdk'
+        run: npm run test-e2e
+        env:
+          STREAM_NAME: ${{secrets.TEST_STREAM_NAME}}
```
vincentsong, line 71: I think we should move TEST_STREAM_NAME and TEST_ACCOUNT_ID to action variables, so we can see the value in the log.  TEST_PUBLISH_TOKEN could be secret. 

And I already set up PUBLISHER_DEMO_ACC_ID(vars) and  PUBLISHER_DEMO_TOKEN(secrets) for actions, you could use it here if you don't need dedicate ones.

Santiago-Souto: Which is the PUBLISHER_DEMO_TOKEN's stream name and account id associated in the variables?

.github/workflows/check-tests.yml
```diff
@@ -57,9 +57,17 @@ jobs:
...
-      - name: Test 🧪
+      - name: Prepare Tests
         working-directory: './packages/millicast-sdk'
-        run: |
-          npm install
-          npx puppeteer browsers install chrome
-          npm run test
+        run: npm ci
+
+      - name: Unit Testing 🧪
```
vincentsong, line 64: We could separate Unit Test and e2e Test in different jobs which will be faster I guess

PR #314, opened 2024-02-05
Feature/integrate netlify

vincentsong:
- Integrated Netlify CLI
- Added setup node & npm Install action for all workflow
- Added deployment for Release Candidate
    - Generated JS Doc
    - Multiview Demo App
    - Publisher Demo App


.github/actions/setup-install/action.yml
```diff
@@ -0,0 +1,16 @@
+name: Setup and Install
+description: Setup nodejs and install npm packages
+runs:
+  using: 'composite'
+  steps:
+    - name: Setup Node.js and Cache
+      uses: actions/setup-node@v4
+      with:
+        node-version: 18
```
Santiago-Souto, line 9: Could we bump it to version 20?

Santiago-Souto (dismissed):
Looking great! Apart from the comment let's go for it?

.github/actions/setup-install/action.yml
vincentsong: Done. Let's see if it is stable enough,  since it's too new.

.github/workflows/rc-release.yml
```diff
@@ -0,0 +1,110 @@
+name: Release Candidate
+
+on:
+  push:
+    branches:
+      - 'RC-[0-9]+.[0-9]+.[0-9]+'
+    tags:
+      - 'next-*'
```
Santiago-Souto, line 8: Isn't it better to trigger with a PR to main from these branch?

vincentsong: When we tag a version like 'next-1.0.2-rc-1', the publish job will start.

PR #320, opened 2024-02-09
Fix incorrect stats count on diagnose call

Santiago-Souto:


packages/millicast-sdk/src/utils/Diagnostics.js
```diff
@@ -11,6 +11,7 @@ const diagnostics = {
   connection: '',
   stats: []
 }
+const tempStats = []
```
vincentsong, line 14: We don't need this actually, diagnostics.stats should be enough

packages/millicast-sdk/src/utils/Diagnostics.js
```diff
@@ -26,10 +27,10 @@ const setFeedId = (feedId) => { diagnostics.feedId = feedId }
 const setConnectionState = (connectionState) => { diagnostics.connection = connectionState }
 
 const setStats = (stats) => {
-  if (diagnostics.stats.length === MAX_STATS_HISTORY_SIZE) {
-    diagnostics.stats.shift()
+  if (tempStats.length === MAX_STATS_HISTORY_SIZE) {
```
vincentsong, line 30:
```suggestion
  if (tempStats.length === Math.min(statsCount, MAX_STATS_HISTORY_SIZE)) {
```

packages/millicast-sdk/src/utils/Diagnostics.js
```diff
@@ -26,10 +27,10 @@ const setFeedId = (feedId) => { diagnostics.feedId = feedId }
...
+    tempStats.shift()
   }
-  diagnostics.stats.push(stats)
+  tempStats.push(stats)
 }
 
 const getDiagnostics = (statsCount = MAX_STATS_HISTORY_SIZE) => {
```
vincentsong, line 36: We also need to check negtive statsCount like below:
```suggestion
if (statsCount > MAX_STATS_HISTORY_SIZE || statsCount <= 0)
```

packages/millicast-sdk/src/utils/Diagnostics.js
Santiago-Souto: `statsCount` cannot be access from here. This is a method called from the PeerConnectionStats interval on the `initStats` function. The temp array seems to be the only way to go.

packages/millicast-sdk/src/utils/Diagnostics.js
Santiago-Souto: Why are you removing the main method?

vincentsong: Sorry, it's not removing the signature, should be the next line, I will update it

packages/millicast-sdk/src/utils/Diagnostics.js
vincentsong: OK, I got it. `statsCount` is only for `getStats` function. e.g. Although we could record 5 at most, user still could get 3 of the latest ones. You can ignore that suggestion. But I am still thinking `diagnostics.stats` should work, isn't it?

Santiago-Souto: I thought it did work, but some times, the interval were the stats are being saved is called during the getDiagnostics function and it overrides any slice or change done before

packages/millicast-sdk/src/utils/Diagnostics.js
```diff
@@ -39,9 +40,14 @@ const getDiagnostics = (statsCount = MAX_STATS_HISTORY_SIZE) => {
   diagnostics.version = version
```
vincentsong, line 40: I am thinking we should not return the global diagnostics object, instead should return a new object in this function when user invoke it. Otherwise, if user wants to keep two diagnostics records, the second function call will modify the first one.

packages/millicast-sdk/src/utils/Diagnostics.js
```diff
@@ -39,9 +40,14 @@ const getDiagnostics = (statsCount = MAX_STATS_HISTORY_SIZE) => {
...
   diagnostics.timestamp = Date.now()
   diagnostics.userAgent = window?.navigator?.userAgent || 'No user agent available'
+  diagnostics.stats = [...tempStats]
 
-  if (diagnostics.stats.length && Number.isInteger(statsCount) && statsCount >= 0 && statsCount < MAX_STATS_HISTORY_SIZE) {
-    diagnostics.stats = diagnostics.stats.slice(-statsCount)
+  if (tempStats.length && Number.isInteger(statsCount) && statsCount >= 0 && statsCount < MAX_STATS_HISTORY_SIZE) {
```
vincentsong, line 45: If the condition is failed, we still return the result? The stats could still hold the old records.

Santiago-Souto: Would it be OK to reset the stats?

vincentsong: No, we just provide the latest stats, it's up to the user if they want to save them or not.

vincentsong: we have checked the statsCount, if it's invalid it will be MAX_STATS_HISTORY_SIZE, isn't it? And I think we should return the latest stats which meet the required statsCount, that's it.

packages/millicast-sdk/src/utils/Diagnostics.js
```diff
@@ -39,9 +40,14 @@ const getDiagnostics = (statsCount = MAX_STATS_HISTORY_SIZE) => {
...
+    if (statsCount === 0) {
+      diagnostics.stats = []
+    } else {
+      diagnostics.stats = tempStats.slice(-statsCount)
```
vincentsong, line 49:
```suggestion
      diagnostics.stats = tempStats.slice(Math.max(-statsCount, -MAX_STATS_HISTORY_SIZE))
```

Santiago-Souto: This would be resolved with the first check of `statsCount`

packages/millicast-sdk/src/utils/Diagnostics.js
vincentsong, line 14: Please refactor the diagnostics object, as we discussed, some information are constants such as version, accountId, streamName, subscriberID, userAgent.   Since we return the copy of `tempStats`, we could just name it `stats`

packages/millicast-sdk/src/utils/Diagnostics.js
```diff
@@ -1,60 +1,52 @@
...
 
 const Diagnostics = {
-  setAccountId,
-  setStreamNameAndSubscriberId,
-  setStreamViewId,
-  setFeedId,
-  setConnectionState,
-  setStats,
-  get: getDiagnostics
+  setAccountId: (newAccountId) => { accountId = newAccountId },
```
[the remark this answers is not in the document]
Santiago-Souto: The problem is that most of these variables are set from different parts of the code, so not possible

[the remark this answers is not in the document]
vincentsong: If we cannot a one init function for all of these, we could have initXYZ instead of setXYZ for each, which could It can imply to the user that these variables can only be initialized once and cannot be changed afterward

vincentsong (approved):

PR #334, opened 2024-03-26
Exposing sendMetadata & worker method

Santiago-Souto:


packages/millicast-sdk/src/workers/TransformWorker.js
```diff
@@ -52,6 +58,10 @@ addEventListener('message', (event) => {
     case 'insertable-streams-receiver':
       setupPipe(event.data, createReceiverTransform())
       break
+    case 'metadata':
```
vincentsong, line 61: Advice to use `metadata-sei-user-data-unregistered` as message action name

packages/millicast-sdk/src/utils/Codecs.js
```diff
@@ -193,10 +193,91 @@ export function extractH26xMetadata (encodedFrame, codec) {
...
+}
+
 export function addH26xSEI ({ uuid, payload }, encodedFrame, codec) {
   if (codec !== 'h264' && codec !== 'h265') {
     throw new Error(`Unsupported codec ${codec}`)
   }
-  // TODO: create new Uint8Array as NAL unit. NAL unit format is start code + header + sei type + payload then push to the original NAL units array
-  return encodedFrame
+  if (uuid === '' || payload === '') {
+    return
```
vincentsong, line 266: We should throw error in this function instead of return which is not good for developer to debug
```suggestion
    throw new Error('uuid and payload cannot be empty')
```

packages/millicast-sdk/src/workers/TransformWorker.js
```diff
@@ -19,6 +22,9 @@ function createSenderTransform () {
     start () {},
     flush () {},
     async transform (encodedFrame, controller) {
+      addH26xSEI({ uuid, payload }, encodedFrame, 'h264')
```
vincentsong, line 25:
```suggestion
    if (uuid && payload) {
      addH26xSEI({ uuid, payload }, encodedFrame, 'h264')
```

packages/millicast-sdk/src/workers/TransformWorker.js
```diff
@@ -52,6 +58,10 @@ addEventListener('message', (event) => {
     case 'insertable-streams-receiver':
       setupPipe(event.data, createReceiverTransform())
       break
+    case 'metadata-sei-user-data-unregistered':
+      uuid = event.data.uuid
```
vincentsong, line 62: We also need to support how to insert the data, e.g.  insert once/ insert X times / every frame / ....
Current mode is every frame, isn't it?

Santiago-Souto: In this case would be each time the function is called

vincentsong:
@Santiago-Souto  please update your branch and fix any bugs, then we could merge it

R-Delfino95 (approved):

PR #336, opened 2024-04-02
Change simulcast docs to enable in all Chromium browsers

Santiago-Souto:


dubeyShivank:
Question - do we need to update `UserAgent.js` and update the check in there to allow edge/chromium based browsers etc?

Santiago-Souto:
> Question - do we need to update `UserAgent.js` and update the check in there to allow edge/chromium based browsers etc?

@dubeyShivank You are right, I just added a new method to check if the UA has `Chrome` in it, as that should imply that is a Chromium based browser.

dubeyShivank (approved):

PR #339, opened 2024-04-10
Enhancements to Diagnostics

justokman:
- Stats initialized implicitly with deprecation warnings for calls to init(), initStats()
- Supplanting local stats parsing with @dolbyio/webrtc-stats library
- Diagnose.get() now accepts a config object with the addition of 'format' for CMCD compatible data
- Changing default stats array length to 60 for 1 minutes data

### DIFF
--- a/packages/millicast-sdk/package-lock.json
+++ b/packages/millicast-sdk/package-lock.json
@@ -1,14 +1,15 @@
 {
   "name": "@millicast/sdk",
-  "version": "0.1.44-rc-1",
+  "version": "0.1.44",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@millicast/sdk",
-      "version": "0.1.44-rc-1",
+      "version": "0.1.44",
       "license": "See in LICENSE file",
       "dependencies": {
+        "@dolbyio/webrtc-stats": "^0.4.0",
         "@types/node": "^18.11.10",
         "Base64": "^1.1.0",
         "buffer": "^6.0.3",
@@ -1899,19 +1900,12 @@
         "node": ">=0.1.95"
       }
     },
-    "node_modules/@hapi/hoek": {
-      "version": "9.3.0",
-      "resolved": "https://registry.npmjs.org/@hapi/hoek/-/hoek-9.3.0.tgz",
-      "integrity": "sha512-/c6rf4UJlmHlC9b5BaNvzAcFv7HZ2QHaV0D4/HNlBdvFnvQq8RI4kYdhyPCl7Xj+oWvTWQ8ujhqS53LIgAe6KQ==",
-      "dev": true
-    },
-    "node_modules/@hapi/topo": {
-      "version": "5.1.0",
-      "resolved": "https://registry.npmjs.org/@hapi/topo/-/topo-5.1.0.tgz",
-      "integrity": "sha512-foQZKJig7Ob0BMAYBfcJk8d77QtOe7Wo4ox7ff1lQYoNNAb6jwcY1ncdoy2e9wQZzvNy7ODZCYJkK8kzmcAnAg==",
-      "dev": true,
+    "node_modules/@dolbyio/webrtc-stats": {
+      "version": "0.4.0",
+      "resolved": "https://registry.npmjs.org/@dolbyio/webrtc-stats/-/webrtc-stats-0.4.0.tgz",
+      "integrity": "sha512-u4OrGbhUn4LM3Ihtoexye+q6eT2jcHsrwVkhFV28DqGLOmy5HJ/5Rt1eoy2+GPErYFWJ0ezLAw8suU5ZBJK4Vw==",
       "dependencies": {
-        "@hapi/hoek": "^9.0.0"
+        "js-logger": "^1.6.1"
       }
     },
     "node_modules/@isaacs/cliui": {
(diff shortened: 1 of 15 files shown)


packages/millicast-sdk/src/PeerConnectionStats.js
```diff
@@ -62,221 +63,62 @@ export const peerConnectionStatsEvents = {
...
+      })
+      this.collection.on('stats', (stats) => {
+        Diagnostics.addStats(stats)
+        this.emit(peerConnectionStatsEvents.stats, stats)
+      })
+      this.collection.start()
+    } catch (e) {
+      logger.error(e)
+    }
+    this.initialized = true
```
dubeyShivank, line 103: should initialized be set before `this.collection.start()` ?

packages/millicast-sdk/src/PeerConnectionStats.js
```diff
@@ -62,221 +63,62 @@ export const peerConnectionStatsEvents = {
...
   }
 
   /**
    * Parse incoming RTCPeerConnection stats.
+   * @deprecated since version 0.1.45 - will be removed in future releases.
    * @param {RTCStatsReport} rawStats - RTCPeerConnection stats.
-   * @returns {ConnectionStats} RTCPeerConnection stats parsed.
+   * @returns {null} Method deprecated and no longer returns meaningful data.
    */
   parseStats (rawStats) {
```
dubeyShivank, line 112: For customers who use the SDK via CDN and not NPM, this might result in a breaking change. 

I like the warning, but we might have to keep this method around for a bit. 

@j12y @vincentsong what do you guys think?

Santiago-Souto:
@justokman please make sure to bring `main` branch as #340 was merged right now, and could bring some conflicts

packages/millicast-sdk/src/PeerConnectionStats.js
j12y: I may need some clarification on scope to answer, because if this is only used in the results of `diagnose()` then that is fine, we want `diagnose()` to return webrtc-stats library results and keys.

If this is a change to both `diagnose()` and the results of `webRTCPeer.on('stats', ...)` then that is more than we want to change at this time. There would be dependencies on hosted player, etc. we would need to reconcile for that deep of a change.

in other words:
- if diagnose is all that is impacted, this is fine
- if all `initStats()` users are impacted, then that's a bigger deal we should revisit

justokman: @dubeyShivank It looks like this method never returned an object like that. 

It is used to parse the raw stats in the original interval and was exposed publicly.

I think it will be safe to deprecate it.

packages/millicast-sdk/src/PeerConnectionStats.js
justokman: Moving to try block.

packages/millicast-sdk/src/PeerConnectionStats.js
dubeyShivank: Ok sweet - thanks for confirming!

packages/millicast-sdk/package-lock.json
```diff
@@ -1,14 +1,16 @@
...
   "requires": true,
   "packages": {
     "": {
       "name": "@millicast/sdk",
-      "version": "0.1.44-rc-1",
+      "version": "0.1.44",
       "license": "See in LICENSE file",
       "dependencies": {
+        "@dolbyio/webrtc-stats": "^0.4.0",
+        "@peermetrics/webrtc-stats": "^5.7.1",
```
dubeyShivank, line 13: Just wanting to be sure - do we need `peermetrics` as well?

packages/millicast-sdk/src/PeerConnection.js
```diff
@@ -43,11 +43,15 @@ export default class PeerConnection extends EventEmitter {
   /**
    * Instance new RTCPeerConnection.
    * @param {RTCConfiguration} config - Peer configuration.
+   * @param {Boolean} [autoInitStats = true] - True to initialize statistics monitoring of the RTCPeerConnection accessed via Logger.get(), false to opt-out.
    */
-  async createRTCPeer (config = null) {
+  async createRTCPeer (config = { autoInitStats: true }) {
```
dubeyShivank, line 48: Probably worth checking - I wasn't seeing stats by default in `Logger.diagnose()` on the publishing side. maybe my setup was fudged, I dunno - but would be good to cross-check

packages/millicast-sdk/package-lock.json
justokman: Huh, I thought I removed that.  Not sure how it got there.

packages/millicast-sdk/src/PeerConnection.js
justokman: Logger.diagnose() works as expected for me after starting a stream in the publisher demo.  We can sync up at standup on this.

packages/millicast-sdk/src/utils/Diagnostics.js
```diff
@@ -10,6 +11,36 @@ let _feedId = ''
...
+    }
+  }
+
+  diagnostics.stats = diagnostics.stats.reduce((acc, stat) => {
+    const audioStats = stat.audio.inbounds.length !== 0
+      ? stat.audio.inbounds.map(statAudio => mapStats('audio', statAudio))
+      : stat.audio.outbounds.map(statAudio => mapStats('audio', statAudio))
+    const videoStats = stat.video.inbounds.length !== 0
+      ? stat.video.inbounds.map(statVideo => mapStats('video', statVideo))
+      : stat.video.outbounds.map(statVideo => mapStats('video', statVideo))
```
Santiago-Souto, line 36: @dubeyShivank In refactored this lines as they were only considering inbound stats, so when testing publish side CMCD stats where not shown. Please review it as well.

dubeyShivank: Thanks Santi, good call out. I will address it

dubeyShivank (dismissed):

Santiago-Souto:
Forced push commits to ensure signing. Will move on to merging this PR

Santiago-Souto (approved):

PR #343, opened 2024-04-19
Throw error when simulcast cannot be set due to compatibility

Santiago-Souto:
@dubeyShivank / @j12y  we should review if you should throw an error and prevent from connecting or log the error and continue without setting simulcast.


Santiago-Souto:
Branched out from wrong branch, closing this PR

PR #344, opened 2024-04-19
Throw error when simulcast cannot be set due to compatibility

Santiago-Souto:
@dubeyShivank / @j12y we should review if you should throw an error and prevent from connecting or log the error and continue without setting simulcast.

### DIFF
--- a/packages/millicast-sdk/src/utils/SdpParser.js
+++ b/packages/millicast-sdk/src/utils/SdpParser.js
@@ -41,12 +41,10 @@ const SdpParser = {
     logger.info('Setting simulcast. Codec: ', codec)
     const browserData = new UserAgent()
     if (!browserData.isChromium()) {
-      logger.warn('Simulcast is only available in Chromium based browser')
-      return sdp
+      throw new Error('Simulcast is only available in Chromium based browser')
     }
     if (codec !== 'h264' && codec !== 'vp8') {
-      logger.warn('Simulcast is only available in h264 and vp8 codecs')
-      return sdp
+      throw new Error('Simulcast is only available in h264 and vp8 codecs')
     }
 
     try {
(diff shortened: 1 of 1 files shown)


packages/millicast-sdk/src/utils/SdpParser.js
```diff
@@ -41,12 +41,10 @@ const SdpParser = {
     logger.info('Setting simulcast. Codec: ', codec)
     const browserData = new UserAgent()
     if (!browserData.isChromium()) {
-      logger.warn('Simulcast is only available in Chromium based browser')
-      return sdp
+      throw new Error('Simulcast is only available in Chromium based browser')
```
dubeyShivank, line 44: @j12y what do you think? I was ok with a descriptive enough warning.

Santiago-Souto: If the warning is the way to go, we could add something like `simulcast not enabled because is only available...`

Santiago-Souto:
@j12y I also changed the warning for codecs to match your proposal of browsers, please have a look at it as well

packages/millicast-sdk/src/utils/SdpParser.js
```diff
@@ -41,11 +41,11 @@ const SdpParser = {
     logger.info('Setting simulcast. Codec: ', codec)
     const browserData = new UserAgent()
     if (!browserData.isChromium()) {
-      logger.warn('Simulcast is only available in Chromium based browser')
+      logger.warn('Your browser does not appear to support Simulcast. For a better experience, use a Chromium based browser.')
       return sdp
     }
     if (codec !== 'h264' && codec !== 'vp8') {
-      logger.warn('Simulcast is only available in h264 and vp8 codecs')
+      logger.warn('Your selected codec does not appear to support Simulcast. For a better experience, use H.264 and VP8 codecs.')
```
dubeyShivank, line 48:
```suggestion
      logger.warn(`Your selected codec ${codec} does not appear to support Simulcast.  To broadcast using simulcast, please use H.264 or VP8.`)
```

dubeyShivank:
LGTM :)

dubeyShivank (approved):

PR #345, opened 2024-04-21
Typescript definitions for diagnose

dubeyShivank:
@Santiago-Souto hoping this is what I need to be doing for ts support?

### DIFF
--- a/.changeset/itchy-walls-burn.md
+++ b/.changeset/itchy-walls-burn.md
@@ -1,0 +1,5 @@
+---
+"@millicast/sdk": patch
+---
+
+Add TS definitions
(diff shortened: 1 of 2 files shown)


Santiago-Souto:
It is the way to go yes! I just do not think the changeset should the created here, as the feature/change is in the `stats` branch, and this is just an add-on to that

dubeyShivank:
Ok I can remove the changeset, thanks!

Santiago-Souto:
Should we create a type for the config object?

Santiago-Souto:
Should `PeerConnectionStats` be updated as well?

dubeyShivank:
> Should `PeerConnectionStats` be updated as well?

I'll do both, thanks!

Santiago-Souto (approved):
LGTM

PR #349, opened 2024-05-17
Don't enable encodedInsertableStreams for non H264 streams

dubeyShivank:


packages/millicast-sdk/src/Publish.js
```diff
@@ -128,7 +127,19 @@ export default class Publish extends BaseWebRTC {
...
+      ...options,
+      peerConfig: {
+        ...connectOptions.peerConfig,
+        ...options.peerConfig
+      },
+      setSDPToPeer: false
+    }
+    this.options.peerConfig.encodedInsertableStreams = this.options.codec === VideoCodec.H264
+
+    logger.debug('this.options', this.options)
```
vincentsong, line 142: Do we need to add this?

dubeyShivank: Oops, sorry! Thanks for pointing out :)

PR #350, opened 2024-05-17
Avoid metadata extracion on non-h264 codec stream

Santiago-Souto:


packages/millicast-sdk/src/utils/Codecs.js
```diff
@@ -336,9 +337,10 @@ function getSeiNalus (frameBuffer, codec) {
         case NALUType.SLICE_PARTITION_A:
           try {
             spsState.findActiveSPS(removePreventionBytes(nalu.subarray(startCodeLength + headerLength)))
-            shouldSearchActiveSPS = false
           } catch (err) {
             console.warn('Failed to find active SPS. Will not be able to extract PIC timing metadata')
+          } finally {
+            shouldSearchActiveSPS = false
```
dubeyShivank, line 343: @vincentsong @mildsunrise any challenges you foresee with this approach of not searching for active SPS more than once?

vincentsong: I prefer to have a flag for parsing pic_timing SEI or not. If we need to parse pic_timing SEI we should keep looking for the active SPS, otherwise we don't need the spsState and also no need to search it.

PR #351, opened 2024-05-21
Prevent metadata SEI extraction from codec - payload type mapping

Santiago-Souto:
Should close #350 and replace it with this one.

Created an object in View instance to gather the SDP mapping of codec - payload type so that this is saved in the worker, and looked at when an encoded frame comes in.


dubeyShivank (approved):
LGTM. Lets RC this!

PR #352, opened 2024-05-23
Check mid before emitting track event

vincentsong:
We need to check mid and add the tracks to the stream before emitting track event


vincentsong:
We have solved this issue in PR #363

PR #355, opened 2024-05-23
Fix / All sources not shown when multiview enabled

Santiago-Souto:
~Change `this.worker` variable for `this.workers` object in View js for handling multiple source streams. Each key correspond to its mid value.~
In order for the View instance to handle multiple source streams tracks (video and audio), every incoming track is used to create an encoded stream and passed it through the worker to process it. To maintain knowledge of which track is which, the transceiver's mid value is saved with the encoded stream created.

Additionally, add publisher demo `sourceId` query param for multisource publishing and expose `millicastView` in multiview demo

### DIFF
--- a/packages/millicast-sdk/src/Publish.js
+++ b/packages/millicast-sdk/src/Publish.js
@@ -297,12 +297,22 @@ export default class Publish extends BaseWebRTC {
    * @param {String} [uuid="6e9cfd2a-5907-49ff-b363-8978a6e8340e"] String with UUID format as hex digit (XXXX-XX-XX-XX-XXXXXX).
    */
   sendMetadata (message, uuid = DOLBY_SEI_DATA_UUID) {
-    if (this.worker) {
+    if (this.worker && this.options.codec === VideoCodec.H264 && !this.options.disableVideo) {
       this.worker.postMessage({
         action: 'metadata-sei-user-data-unregistered',
         uuid: uuid,
         payload: message
       })
+    } else {
+      let warningMessage = 'Could not send metadata. Reason: '
+      if (this.options.codec !== VideoCodec.H264) {
+        warningMessage += 'Incompatible codec. Only H264 available.'
+      } else if (this.options.disableVideo) {
+        warningMessage += 'Video disabled.'
+      } else {
+        warningMessage += 'Unknown.'
+      }
+      logger.warn(warningMessage)
     }
   }
 };
--- a/packages/millicast-sdk/src/Publish.js
+++ b/packages/millicast-sdk/src/Publish.js
@@ -169,6 +169,11 @@ export default class Publish extends BaseWebRTC {
     }
   }
 
+  stop () {
+    super.stop()
+    this.worker?.terminate()
+  }
+
   async initConnection (data) {
     logger.debug('Broadcast option values: ', this.options)
     this.stopReconnection = false
@@ -232,26 +237,27 @@ export default class Publish extends BaseWebRTC {
 
     const workerBlob = new Blob([workerString])
     const workerURL = URL.createObjectURL(workerBlob)
-    const worker = new Worker(workerURL)
+    this.worker = new Worker(workerURL)
 
-    const senders = this.getRTCPeerConnection()
-      .getSenders()
-      .filter((elt) => elt.track.kind === 'video')
-    const sender = senders[0]
+    const senders = this.getRTCPeerConnection().getSenders()
 
-    if (supportsRTCRtpScriptTransform) {
-      // eslint-disable-next-line no-undef
-      sender.transform = new RTCRtpScriptTransform(worker, { name: 'senderTransform', codec: this.options.codec })
-    } else if (supportsInsertableStreams) {
-      const { readable, writable } = sender.createEncodedStreams()
-      worker.postMessage({
-        action: 'insertable-streams-sender',
-        codec: this.options.codec,
-        readable,
-        writable
-      }, [readable, writable])
-    }
-    this.worker = worker
+    senders.forEach(sender => {
+      if (supportsRTCRtpScriptTransform) {
+        // eslint-disable-next-line no-undef
+        sender.transform = new RTCRtpScriptTransform(this.worker, {
+          name: 'senderTransform',
+          codec: this.options.codec
+        })
+      } else if (supportsInsertableStreams) {
+        const { readable, writable } = sender.createEncodedStreams()
+        this.worker.postMessage({
+          action: 'insertable-streams-sender',
+          codec: this.options.codec,
+          readable,
+          writable
+        }, [readable, writable])
+      }
+    })
 
     let oldSignaling = this.signaling
     this.signaling = signalingInstance
--- a/packages/millicast-sdk/src/Publish.js
+++ b/packages/millicast-sdk/src/Publish.js
@@ -172,6 +172,7 @@ export default class Publish extends BaseWebRTC {
   stop () {
     super.stop()
     this.worker?.terminate()
+    this.worker = null
   }
 
   async initConnection (data) {
--- a/packages/millicast-sdk/src/Publish.js
+++ b/packages/millicast-sdk/src/Publish.js
@@ -238,6 +238,9 @@ export default class Publish extends BaseWebRTC {
 
     const workerBlob = new Blob([workerString])
     const workerURL = URL.createObjectURL(workerBlob)
+    if (this.worker) {
+      this.worker.terminate()
+    }
     this.worker = new Worker(workerURL)
 
     const senders = this.getRTCPeerConnection().getSenders()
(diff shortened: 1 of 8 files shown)


packages/millicast-sdk/src/Publish.js
```diff
@@ -169,6 +169,11 @@ export default class Publish extends BaseWebRTC {
     }
   }
 
+  stop () {
+    super.stop()
+    this.worker?.terminate()
```
dubeyShivank, line 174: Do we need to do this?
```suggestion
    this.worker?.terminate()
    this.worker = null
```

packages/millicast-sdk/src/Publish.js
```diff
@@ -232,26 +237,27 @@ export default class Publish extends BaseWebRTC {
...
     const workerBlob = new Blob([workerString])
     const workerURL = URL.createObjectURL(workerBlob)
-    const worker = new Worker(workerURL)
+    this.worker = new Worker(workerURL)
 
-    const senders = this.getRTCPeerConnection()
-      .getSenders()
-      .filter((elt) => elt.track.kind === 'video')
-    const sender = senders[0]
+    const senders = this.getRTCPeerConnection().getSenders()
```
dubeyShivank, line 242: What about audio tracks? Do we not need to filter them out anymore?

dubeyShivank: Oh is that because of the no audio bug reported on slack?

Santiago-Souto: Exactly, we need to create all tracks into encoded streams to be seen/heard

dubeyShivank (approved):

packages/millicast-sdk/src/Publish.js
```diff
@@ -232,26 +238,27 @@ export default class Publish extends BaseWebRTC {
 
     const workerBlob = new Blob([workerString])
     const workerURL = URL.createObjectURL(workerBlob)
-    const worker = new Worker(workerURL)
+    this.worker = new Worker(workerURL)
```
vincentsong, line 241: If we cannot ensure this.worker has been terminated here, should we check and terminate it?

packages/millicast-sdk/src/Publish.js
```diff
@@ -290,12 +297,22 @@ export default class Publish extends BaseWebRTC {
...
       this.worker.postMessage({
         action: 'metadata-sei-user-data-unregistered',
         uuid: uuid,
         payload: message
       })
+    } else {
+      let warningMessage = 'Could not send metadata. Reason: '
+      if (this.options.codec !== VideoCodec.H264) {
+        warningMessage += 'Incompatible codec. Only H264 available.'
+      } else if (this.options.disableVideo) {
```
vincentsong, line 310: maybe we need to use `if` to cover all of the cases here

packages/millicast-sdk/src/View.js
```diff
@@ -192,6 +193,9 @@ export default class View extends BaseWebRTC {
   stop () {
     super.stop()
     this.worker?.terminate()
+    this.worker = null
+    this.codecPayloadTypeMap = {}
```
vincentsong, line 197: if it's map, should use `new Map()` otherwise it's an object here

packages/millicast-sdk/src/View.js
```diff
@@ -243,27 +247,34 @@ export default class View extends BaseWebRTC {
 
     const workerBlob = new Blob([workerString])
     const workerURL = URL.createObjectURL(workerBlob)
+    this.worker = new Worker(workerURL)
```
vincentsong, line 250: Please ensure we have terminated the previous worker

packages/millicast-sdk/src/View.js
```diff
@@ -243,27 +247,34 @@ export default class View extends BaseWebRTC {
...
 
     webRTCPeerInstance.on('track', (trackEvent) => {
-      if (trackEvent.track?.kind !== 'video') return
-      const worker = new Worker(workerURL)
+      this.trackMediaId[trackEvent.transceiver?.mid] = trackEvent.track
```
vincentsong, line 253: Looks like this is also a map between track's media id and the track itself, if yes, we should name it `trackMediaIdMap`, and also add comment to explain the purpose of this object.

packages/millicast-sdk/src/workers/TransformWorker.worker.js
```diff
@@ -1,44 +1,78 @@
 import { addH26xSEI, extractH26xMetadata } from '../utils/Codecs'
 
-let uuid = ''
-let payload = ''
+const DROPPED_SOURCE_TIMEOUT = 2000
+const metadata = []
 let codec = ''
 let codecMap = {}
+const synchronizationSources = {}
```
vincentsong, line 7: Could you please add comment to explain the purpose of the synchronizationSources object?

packages/millicast-sdk/src/workers/TransformWorker.worker.js
```diff
@@ -1,44 +1,78 @@
...
     async transform (encodedFrame, controller) {
-      const frameCodec = codecMap[encodedFrame.getMetadata().payloadType] || codec.toUpperCase()
-      if (frameCodec === 'H264') {
-        const metadata = extractH26xMetadata(encodedFrame, frameCodec)
-        if (metadata.timecode || metadata.unregistered || metadata.seiPicTimingTimeCodeArray?.length > 0) {
-          self.postMessage({ metadata })
+      // eslint-disable-next-line no-undef
+      if (encodedFrame instanceof RTCEncodedVideoFrame) {
+        const frameCodec = codecMap[encodedFrame.getMetadata().payloadType] || codec.toUpperCase()
+        if (frameCodec === 'H264') {
```
vincentsong, line 18: Will `codecMap[encodedFrame.getMetadata().payloadType]` always return uppercase codec string?

packages/millicast-sdk/src/workers/TransformWorker.worker.js
```diff
@@ -1,44 +1,78 @@
...
+    synchronizationSourcesWithMetadata = []
+  }
+}
+
+function refreshSynchronizationSources (newSyncSource) {
+  const now = new Date().getTime()
+  synchronizationSources[newSyncSource] = now
+
+  const sourcesToDelete = Object.keys(synchronizationSources)
+    .filter(source => (now - synchronizationSources[source] > DROPPED_SOURCE_TIMEOUT))
```
vincentsong, line 42: How do we get the time out value as 2 seconds? Any industry standard requirement?

packages/millicast-sdk/src/workers/TransformWorker.worker.js
```diff
@@ -1,44 +1,78 @@
...
+      if (encodedFrame instanceof RTCEncodedVideoFrame) {
+        const frameMetadata = encodedFrame.getMetadata()
+        const newSyncSource = frameMetadata.synchronizationSource
+
+        refreshSynchronizationSources(newSyncSource)
+
+        if (!synchronizationSourcesWithMetadata.includes(newSyncSource) && metadata.length) {
+          try {
+            // Add h265 regex when ready
+            if (!/(h26[4])/.test(codec)) {
```
vincentsong, line 66: how about H264 here (uppercase)?

vincentsong:
> Change `this.worker` variable for `this.workers` object in View js for handling multiple source streams. Each key correspond to its mid value.
> 
> Additionally, add publisher demo `sourceId` query param for multisource publishing and expose `millicastView` in multiview demo
Do you need to update this description? I didn't find `this.workers`, it's still using `this.worker`

packages/millicast-sdk/src/View.js
Santiago-Souto: Will change the name. It works fine as an object, and it is treated as that.

packages/millicast-sdk/src/View.js
Santiago-Souto: As it will be used as an object, I think it is good not to include map in it, will change it to `tracksMidValues` for better clarity

packages/millicast-sdk/src/workers/TransformWorker.worker.js
Santiago-Souto: They come from the SDP, and the cases we are testing all come in uppercase. I can add a `toUpperCase` there as well to ensure both will work

packages/millicast-sdk/src/workers/TransformWorker.worker.js
Santiago-Souto: No standard found, it was a number we thought we could make sure the resolution dropped. Any suggestions are welcome

packages/millicast-sdk/src/workers/TransformWorker.worker.js
Santiago-Souto: This codec comes from the `options.codec` of the connect Publish method, that is set to be lower case

packages/millicast-sdk/src/workers/TransformWorker.worker.js
```diff
@@ -59,9 +95,9 @@ addEventListener('rtctransform', (event) => {
     codec = event.transformer.options.codec
     transform = createSenderTransform()
   } else if (event.transformer.options.name === 'receiverTransform') {
-    codecMap = event.transformer.options.codecMap || {}
+    payloadTypeCodec = event.transformer.options.payloadTypeCodec || {}
```
dubeyShivank, line 98: nitpick non-blocking : `payloadCodecType`

packages/millicast-sdk/src/workers/TransformWorker.worker.js
dubeyShivank: So the reason this function exists is to ensure that the `synchronisationSources` Map doesn't keep growing if sources keep getting added/removed from the list, isn't it? It's to ensure faster lookups

dubeyShivank: I'm wondering if we need to store the timestamp at all, and if it can just be an array of `ssrc`. What's the implication of that. 

@vincentsong what do you think?

dubeyShivank (approved):

packages/millicast-sdk/src/workers/TransformWorker.worker.js
Santiago-Souto: The thing is that is the codec's payload type, so these are two different attributes

PR #357, opened 2024-05-28
Do not use a hardcoded list of capabilities for firefox

dubeyShivank:
Draft until we finalise the current release and then consider merging this into `main`


Santiago-Souto (approved):
Ship it!

PR #363, opened 2024-06-05
Integrate castlab SDK

vincentsong:
Precondition:
1.  Use query parameter '**subscriberToken**' or env var 'MILLICAST_SUBSCRIBER_TOKEN'
2. Use query parameter '**drm**' to toggle DRM mode in URL

To-dos:
- [x] Update workflow of fetching DRM configuration from MediaServer and backend Director API. e.g. KeyId/iv/license server URL
- [x] use media ID as video/audio element id instead of using source id
- [x] Clean up demo apps

Known issues:
1. Only available in `Amsterdam` MediaServer cluster
2. We are still in DRMToday's staging environment, need to change to production environment when it's ready


packages/millicast-sdk/src/View.js
```diff
@@ -323,4 +400,47 @@ export default class View extends BaseWebRTC {
...
+   * @property {DRMMediaOptions} video - video DRM options
+   * @property {DRMMediaOptions} audio -audio DRM options
+   */
+  /**
+   * config DRM protected stream from transceiver
+   * @param {CastLabDRMOptions} options - the DRM options in castlab SDK format
+   * @param {'video' | 'audio'} mediaType - 'video' or 'audio'
+   * @param {String} mediaId - the transceiver's mediaId, 'main' stands for main stream
+   */
+  configDRM (options, mediaType, mediaId) {
```
dubeyShivank, line 426: nitpick : `configureDRM`

packages/millicast-sdk/src/View.js
```diff
@@ -323,4 +400,47 @@ export default class View extends BaseWebRTC {
...
+
+  /**
+   * @typedef {Object} DRMMediaOptions
+   * @property {String} codec - h264 for video
+   * @property {String} encryption - cbsc or clear
+   * @property {String} [keyId] - decryption keyId when encryption is not clear
+   * @property {String} [iv] - fairplay iv when encryption is not clear
+   */
+  /**
+   * @typedef {Object} CastLabDRMOptions extends DRMOptions
```
dubeyShivank, line 412: I'd suggest not using the Castlabs name here, since we might have an in-house implementation of DRM down the line

packages/millicast-sdk/src/View.js
```diff
@@ -207,7 +273,8 @@ export default class View extends BaseWebRTC {
       subscriberData = await this.tokenGenerator()
       //  Set the iceServers from the subscribe data into the peerConfig
       this.options.peerConfig.iceServers = subscriberData?.iceServers
-      this.options.peerConfig.encodedInsertableStreams = supportsInsertableStreams
+      // We should not set the encodedInsertableStreams if the DRM and the frame metadata are not enabled
+      this.options.peerConfig.encodedInsertableStreams = supportsInsertableStreams && (this.isDRMOn || this.isFrameMetadataOn)
```
dubeyShivank, line 277: `this.frameMetadataOn` isn't set anywhere or defined anywhere? It isn't a roadblock for right now but would need cleaning up before we release DRM with the SDK :)

packages/millicast-sdk/src/View.js
vincentsong: But this is for castLab DRM options, we could name our in-house implementation DolbyDRMOptions. If don't use this, what name do you suggest?

packages/millicast-sdk/src/View.js
vincentsong: Thanks, will rename it

packages/millicast-sdk/src/View.js
vincentsong: Yes, this flag is undefined now, and frame metadata is not compatible with DRM. But I am thinking we should have a flag to identify this feature.

dubeyShivank: I completely agree that we should have a flag to identify this option :) Lets keep it, but absolutely, we can clean up and define later :)

packages/millicast-sdk/src/View.js
dubeyShivank: how do you feel about `vendorDRMOptions`

vincentsong: What if there is another vendor?

dubeyShivank: It's generic enough to still be true

dubeyShivank: Resolving it - no legal implications of this, so not worth a long discussion :)

packages/millicast-sdk/src/View.js
```diff
@@ -323,4 +373,104 @@ export default class View extends BaseWebRTC {
...
+   * @property {EncryptionParameters} [audioEncParams] - the audio encryption parameters
+   * @property {String} [audioMid] - the audio media ID of RTCRtpTransceiver
+   */
+
+  /**
+   * Configure DRM protected stream.
+   * When there are {@link EncryptionParameters} in the payload of 'active' broadcast event, this method should be called
+   * @param {DRMOptions} options - the options for DRM playback
+  */
+  configureDRM (options) {
```
dubeyShivank, line 402: Question - should `options` be an object like this (or whatever variable names you like)- 

``` typescript
export interface DRMOptions {
 videoParams : VideoDRMOptions,
 audioParams : AudioDRMOptions
}

export interface VideoDRMOptions {
  videoElement : HTMLVideoElement,
  encryptionParams : EncryptionParameters,
  mid : String 
}

export interface AudioDRMOptions {
    audioElement : HTMLAudioElement,
    encryptionParams : EncryptionParameters,
    mid : String 
}
```

vincentsong: I prefer flat structure which is easier to read and modify since this is not a complex one. Also in some cases, we only need 'audioElement' but no 'encryptionParams' and 'mid' for audio.

dubeyShivank: Disagree but 👍

packages/millicast-sdk/src/PeerConnection.js
```diff
@@ -394,6 +395,10 @@ const instanceRTCPeerConnection = (instanceClass, config) => {
   return instance
 }
 
+async function delay (ms) {
```
dubeyShivank, line 398: Should this be private?

vincentsong: Yes, this should be private, will update it in next commit

packages/millicast-sdk/src/View.js
vincentsong: have updated with 'options.metadata'

packages/millicast-multiview-demo/src/multiviewer.js
```diff
@@ -29,37 +38,29 @@ let sourceTracksMap = {}
...
-})
-
 document.addEventListener('DOMContentLoaded', async () => {
   remoteVideosContainer = document.getElementById('remoteVideos')
   mainVideoElement = document.getElementById('mid-0')
   mainAudioElement = document.getElementById('mid-1')
-  console.log(' page loaded, mainVideoElement is', mainVideoElement)
   try {
     viewer = new View(streamName, tokenGenerator)
+    viewer.on('onMetadata', (metadata) => {
```
dubeyShivank, line 54: nitpick non blocking : The next version of the SDK will not report `onMetadata`. Instead it will only be `metadata`. You can use that instead

packages/millicast-sdk/src/View.js
```diff
@@ -148,7 +150,7 @@ export default class View extends BaseWebRTC {
     logger.info('Connected to streamName: ', this.streamName)
   }
 
-  #hexToUint8Array (hexString) {
+  hexToUint8Array (hexString) {
```
dubeyShivank, line 153: nitpick non-blocking: I think this can be moved to a utility/helper file

packages/millicast-sdk/src/View.js
```diff
@@ -160,11 +162,11 @@ export default class View extends BaseWebRTC {
     return uint8Array
   }
 
-  #swapPropertyValues (obj1, obj2, key) {
+  swapPropertyValues (obj1, obj2, key) {
```
dubeyShivank, line 165: nitpick non blocking - consider moving to some sort of helper/utility

packages/millicast-sdk/src/types/index.d.ts
```diff
@@ -1248,11 +1249,22 @@ declare module '@millicast/sdk' {
...
+    playReadyUrl?: string;
+    widevineUrl?: string;
+    fairPlayUrl?: string;
+  }
+
   export type DirectorResponse = {
     urls: string[];
     jwt: string;
     iceServers: RTCIceServer[];
+    drmObject?: DRMProfile;
```
dubeyShivank, line 1265: nitpick : can we call this `drmProfile` ?

packages/millicast-sdk/src/types/index.d.ts
```diff
@@ -1359,6 +1371,42 @@ declare module '@millicast/sdk' {
...
+
+  /**
+   * The configuration for DRM playback
+   */
+  export type DRMOptions = {
+    /** The video element */
+    videoElement: HTMLVideoElement;
+
+    /** The video encryption parameters */
+    videoEncParams: EncryptionParameters;
```
dubeyShivank, line 1395: nitpick : can we have `videoEncryptionParams` ? 
```suggestion
    videoEncryptionParams: EncryptionParameters;
```

packages/millicast-sdk/src/types/index.d.ts
```diff
@@ -1359,6 +1371,42 @@ declare module '@millicast/sdk' {
...
+
+    /** The video media ID of RTCRtpTransceiver */
+    videoMid: string;
+
+    /** The audio element */
+    audioElement: HTMLAudioElement;
+
+    /** The audio encryption parameters */
+    audioEncParams?: EncryptionParameters;
```
dubeyShivank, line 1404: same here?

packages/millicast-sdk/src/types/index.d.ts
```diff
@@ -1463,6 +1511,28 @@ declare module '@millicast/sdk' {
...
+
+    /**
+     * Check if there are any DRM protected Track
+     */
+    get isDRMOn(): boolean;
+
+    /** Exchange the DRM configuration between two transceivers
+     *  Make sure both of the transceivers have been used for DRM protected streams
+     */
+    exchangeDRMConfiguration (targetMediaId: string, sourceMediaId: string);
```
dubeyShivank, line 1534: Question : is this function used with a multi-source DRM stream?

packages/millicast-multiview-demo/src/multiviewer.js
vincentsong: will change it

packages/millicast-sdk/src/types/index.d.ts
vincentsong: yes

vincentsong: @dubeyShivank do you have any further questions? Or else please resolve this, thx.

vincentsong:
@dubeyShivank Please feel free to refine the changesets content and push it. If there is no other review comments I think we could merge this PR.

dubeyShivank (approved):
LGTM

PR #366, opened 2024-06-07
Only throw an info message for missing SPS

dubeyShivank:
The application developer can't do anything with the warning, and it's confusing for them because 
- they don't know what triggered it
- we don't officially support SEI_PIC_TIMING


Santiago-Souto (approved):
LGTM

PR #368, opened 2024-06-07
Make metadata optional. Not enabled by default

Santiago-Souto:
Metadata is now optional on both sides, viewer and publisher. If connect option `metadata` is set to true, then `encodedInsertableStreams` peer config will be set and a Web Worker will be created only once. 

For Safari users, metadata extraction will be based only on this option, so if the stream published does not have H264 codec, but the viewer set metadata to true, it will try to extract. This replaces the `options.codec = 'h264'` needed previously, to avoid setting to options instead of only metadata one. We assume that if the user sets metadata extraction, is the same as setting the view codec to h264.

### DIFF
--- a/packages/millicast-publisher-demo/src/publisher.js
+++ b/packages/millicast-publisher-demo/src/publisher.js
@@ -158,8 +158,17 @@ document.addEventListener("DOMContentLoaded", async (event) => {
       });
       let priority = parseInt(params.priority);
       const sourceId = params.sourceId
-      priority = isNaN(priority) ? undefined : priority;
-      await millicastPublishUserMedia.connect({ bandwidth, events: events, priority, sourceId })
+      const metadata = params.metadata === "true"
+      const connectOptions = {
+        bandwidth,
+        events,
+        metadata
+      }
+      if (!isNaN(priority)) {
+        connectOptions.priority = priority
+      }
+      if (sourceId) connectOptions.sourceId = sourceId
+      await millicastPublishUserMedia.connect(connectOptions)
       isBroadcasting = true;
       broadcastHandler();
       setUserCount();
--- a/packages/millicast-publisher-demo/src/publisher.js
+++ b/packages/millicast-publisher-demo/src/publisher.js
@@ -156,7 +156,7 @@ document.addEventListener("DOMContentLoaded", async (event) => {
       const params = new Proxy(new URLSearchParams(window.location.search), {
         get: (searchParams, prop) => searchParams.get(prop),
       });
-      let priority = parseInt(params.priority);
+      const priority = parseInt(params.priority);
       const sourceId = params.sourceId
       const metadata = params.metadata === "true"
       const connectOptions = {
--- a/packages/millicast-sdk/src/Publish.js
+++ b/packages/millicast-sdk/src/Publish.js
@@ -16,13 +16,13 @@ const logger = Logger.get('Publish')
 const connectOptions = {
   mediaStream: null,
   bandwidth: 0,
+  metadata: false,
   disableVideo: false,
   disableAudio: false,
   codec: VideoCodec.H264,
   simulcast: false,
   scalabilityMode: null,
   peerConfig: {
-    encodedInsertableStreams: true,
     autoInitStats: true
   }
 }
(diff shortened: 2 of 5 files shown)


packages/millicast-publisher-demo/src/publisher.js
```diff
@@ -158,8 +158,17 @@ document.addEventListener("DOMContentLoaded", async (event) => {
       });
       let priority = parseInt(params.priority);
```
FabienLavocat, line 159: I would change `let priority` to `const priority`

packages/millicast-sdk/src/Publish.js
```diff
@@ -236,32 +245,33 @@ export default class Publish extends BaseWebRTC {
     promises = await Promise.all([getLocalSDPPromise, signalingConnectPromise])
     const localSdp = promises[0]
 
-    const workerBlob = new Blob([workerString])
-    const workerURL = URL.createObjectURL(workerBlob)
-    if (this.worker) {
-      this.worker.terminate()
```
FabienLavocat, line 242: Shouldn't you keep the cleanup of the worker in case it was set previously?

packages/millicast-sdk/src/View.js
```diff
@@ -247,31 +251,36 @@ export default class View extends BaseWebRTC {
...
-          codec: this.options.codec,
-          mid: trackEvent.transceiver?.mid,
-          readable,
-          writable
-        }, [readable, writable])
+    if (this.options.metadata) {
+      const workerBlob = new Blob([workerString])
+      const workerURL = URL.createObjectURL(workerBlob)
+      if (!this.worker) {
+        this.worker = new Worker(workerURL)
```
FabienLavocat, line 258: Should we clean up the `worker` object in case it was set previously?

FabienLavocat (changes requested):
I've added a few comments, went through the code visually.

packages/millicast-sdk/src/View.js
Santiago-Souto: If the worker is set previously, we could reuse it right? In the `stop` method we do the cleanup. If it is thought that the worker should be replaced in those cases we can clean up as it was. This was change to avoid creating a new Worker each time the method is called.

FabienLavocat: If you reconnect to a stream without the need for metadata, then you want to clean up the worker to release the resources.
This method is only called when you connect to a new stream, so spending a little bit more time (few ms) on initialization should not be a big deal IMO.

Santiago-Souto: When reconnecting, it first calls `stop` (so worker terminated) and then starts connection process again, so I don't know if it's necessary to do a terminate there, right?

FabienLavocat: Ok, if the `stop` is called all the time, then we're good with not cleaning up the worker.

packages/millicast-sdk/src/Publish.js
Santiago-Souto: This is the same as in the View comment, I believe

FabienLavocat: Yes same

FabienLavocat (approved):

packages/millicast-sdk/src/Publish.js
```diff
@@ -195,6 +201,9 @@ export default class Publish extends BaseWebRTC {
       publisherData = await this.tokenGenerator()
       //  Set the iceServers from the publish data into the peerConfig
       this.options.peerConfig.iceServers = publisherData?.iceServers
+      if (this.options.metadata) {
+        this.options.peerConfig.encodedInsertableStreams = supportsInsertableStreams
+      }
```
dubeyShivank, line 206:
``` suggestion
this.options.peerConfig.encodedInsertableStreams = this.options.metadata
```

packages/millicast-sdk/src/types/index.d.ts
```diff
@@ -1071,6 +1071,10 @@ declare module '@millicast/sdk' {
      * - True to modify SDP for supporting absolute capture time header extension. Otherwise False.
      */
     absCaptureTime?: boolean;
+    /**
+     * - Enable metadata extraction if stream is compatible.
```
dubeyShivank, line 1075:
```
* - Enable metadata extraction. This feature is only supported with the H.264 codec. 
* ```

File: packages/millicast-sdk/src/types/index.d.ts [hunk 6]
```diff
@@ -1156,6 +1160,10 @@ declare module '@millicast/sdk' {
      * - Broadcast bandwidth. 0 for unlimited.
      */
     bandwidth?: number;
+    /**
+     * - Enable metadata insertion if stream is compatible.
+     */
+    metadata?: boolean;
```
dubeyShivank, line 1166:
``` suggestion
* - Enable metadata insertion. This feature is only supported with the H.264 codec. 
```

PR #369, opened 2024-06-10
Improve build time performance

juanpiquerez:
Significantly reduce our SDK build times and enable additional features such as hot reload.
Currently, building our SDK locally takes 45 to 50 seconds, which is not ideal for efficient development. 
A faster and more efficient solution needed to be found.

**Implementation**

- _Migration to Vite_: We migrated the build system from Rollup to Vite.

- _Vite Configuration_: We configured Vite to meet the requirements of our SDK and take advantage of its rapid build capabilities.

- _Hot Reload_: We enabled and configured the hot reload functionality, significantly improving the development workflow.


dubeyShivank:
Nice work @juanpiquerez !! Just a couple of things - 

* can we also update the readme if required, please?

* One thing I've been contemplating but dont know for sure is - can we have two commands - `npm run start` and `npm run start-all` ? start only builds and runs 3 packages, start-all can be what we currently have? Or do you feel people will miss testing the multi-view component if we do it that way?

dubeyShivank (approved):

dubeyShivank:
Es muy excellente!

PR #377, opened 2024-06-19
update metadata

dubeyShivank:
* The event triggered by the SDK is now `metadata` instead of `onMetadata`
* Now, even SDK generated messages contain a `timecode` property.


packages/millicast-multiview-demo/src/multiviewer.js
```diff
@@ -60,7 +60,7 @@ viewer.on('track', (event) => {
   }
 })
 
-viewer.on('onMetadata', (metadata) => {
+viewer.on('metadata', (metadata) => {
```
j12y, line 63: 👍

packages/millicast-sdk/src/Publish.js
```diff
@@ -310,10 +309,15 @@ export default class Publish extends BaseWebRTC {
    */
   sendMetadata (message, uuid = DOLBY_SEI_DATA_UUID) {
```
Santiago-Souto, line 310: Now that I am seeing, message is set as a string, but it could also be an object, etc.

dubeyShivank: can update the definition

packages/millicast-sdk/src/types/index.d.ts
```diff
@@ -1327,10 +1327,10 @@ declare module '@millicast/sdk' {
     unrecord(): Promise<void>;
     /**
      * Send SEI user unregistered data as part of the frame being streamed. Only available for H.264 codecs.
-     * @param {String} message String with the data to be sent as SEI user unregistered data.
-     * @param {String} [uuid="6e9cfd2a-5907-49ff-b363-8978a6e8340e"] String with UUID format as hex digit (XXXX-XX-XX-XX-XXXXXX).
+     * @param {String | object} message The data to be sent as SEI user unregistered data.
```
Santiago-Souto, line 1330:
```suggestion
     * @param {String | Object} message The data to be sent as SEI user unregistered data.
```
nitpick

packages/millicast-sdk/src/utils/Codecs.js
```diff
@@ -48,8 +48,18 @@ const SEI_Payload_Type = {
   USER_DATA_UNREGISTERED: 5
 }
 
+const UNREGISTERED_MESSAGE_TYPE = {
+  LEGACY: 1,
+  NEW: 2,
+  TIMECODE: 3
+}
+
+// Old UUID - used by SDK 0.1.45 and before. No longer actively used (only for backwards compatibility purposes)
```
Santiago-Souto, line 57:
```suggestion
// Old UUID - used by SDK 0.1.46 and before. No longer actively used (only for backwards compatibility purposes)
```
It's 46, right?

packages/millicast-sdk/src/utils/Codecs.js
```diff
@@ -372,24 +382,50 @@ function extractSEIPayload (rbsp) {
...
-    metadata.unregistered = content
+  switch (messageType) {
+    case UNREGISTERED_MESSAGE_TYPE.LEGACY :
+      metadata.unregistered = content
+      break
+    case UNREGISTERED_MESSAGE_TYPE.TIMECODE:
+      metadata.timecode = convertSEITimestamp(content)
+      break
+    case UNREGISTERED_MESSAGE_TYPE.NEW:
+    {
```
Santiago-Souto, line 414:
```suggestion
```
Do we need this bracket?

packages/millicast-sdk/src/utils/Codecs.js
```diff
@@ -372,24 +382,50 @@ function extractSEIPayload (rbsp) {
...
+      // Do not make assumptions on the size of a date object as a byte array
+      let index = 0
+      const timecodeBufferLength = numberToByteArray(Date.now()).length
+      const timecodeSubArray = content.subarray(index, timecodeBufferLength)
+      index += timecodeBufferLength
+      const metadataSubArray = content.subarray(index)
+      metadata.timecode = convertSEITimestamp(timecodeSubArray)
+      metadata.unregistered = metadataSubArray
+      break
+    }
```
Santiago-Souto, line 425:
```suggestion
```
Same here

Santiago-Souto (changes requested):
It looks great! Just to make sure the following is intended, in case the user passes a different UUID, the timecode attribute will be ignored.

packages/millicast-sdk/src/utils/Codecs.js
dubeyShivank: Oh I didn't have them, but lint didn't like em, so added them back in

dubeyShivank:
> It looks great! Just to make sure the following is intended, in case the user passes a different UUID, the timecode attribute will be ignored.

Yes - I've covered that in the release notes.

Santiago-Souto (approved):
Go ahead!! 🚀

PR #378, opened 2024-06-24
Allow the stats timeout to be user-configurable

dubeyShivank:


packages/millicast-sdk/src/PeerConnection.js
```diff
@@ -43,13 +43,13 @@ export default class PeerConnection extends EventEmitter {
...
    */
-  async createRTCPeer (config = { autoInitStats: true }) {
-    logger.info('Creating new RTCPeerConnection')
-    logger.debug('RTC configuration provided by user: ', config)
+  async createRTCPeer (config = { autoInitStats: true, statsIntervalMs: 1000 }) {
+    logger.info('Creating new RTCPeerConnection', config)
     this.peer = instanceRTCPeerConnection(this, config)
     if (config.autoInitStats) {
-      this.initStats()
+      this.initStats(config.autoInitStats, config.statsIntervalMs)
```
Santiago-Souto, line 52: Couldn't we pass in config object and then extract only the relevant options. Mainly thinking about scalability

packages/millicast-sdk/src/PeerConnection.js
```diff
@@ -336,11 +336,11 @@ export default class PeerConnection extends EventEmitter {
    *   console.log('Stats from event: ', stats)
    * })
    */
-  initStats () {
+  initStats (autoInitStats, statsIntervalMs) {
```
Santiago-Souto, line 339: Same here, could take something like `options`

packages/millicast-sdk/src/PeerConnection.js
```diff
@@ -336,11 +336,11 @@ export default class PeerConnection extends EventEmitter {
...
     if (this.peerConnectionStats) {
       logger.warn('PeerConnection.initStats() has already been called.  Automatic initialization occurs via View.connect(), Publish.connect() or this.createRTCPeer(). See options')
     } else if (this.peer) {
-      this.peerConnectionStats = new PeerConnectionStats(this.peer)
+      this.peerConnectionStats = new PeerConnectionStats(this.peer, { autoInitStats, statsIntervalMs })
```
Santiago-Souto, line 343: `options` as well, and in the constructor extract it

packages/millicast-sdk/src/PeerConnection.js
dubeyShivank: Yes we can and now we do :)

packages/millicast-sdk/src/PeerConnection.js
dubeyShivank: Yep

packages/millicast-sdk/src/PeerConnection.js
dubeyShivank: Si senor

Santiago-Souto:
@dubeyShivank are we going to save both changeset files? They look similar

dubeyShivank:
> @dubeyShivank are we going to save both changeset files? They look similar

oh, I'll delete one! Thank you! I saw the `no changesets detected` message so I added a new one today. Mea culpa!

Santiago-Souto (approved):
lgtm

PR #392, opened 2024-07-08
WIP: CTA608(EIA608) caption PoC

vincentsong:
This is a PoC of parsing and displaying CTA-608(EIA608) closed caption

- [x] Extracting CC from h264 SEI user data registered
- [x] Parsing and decoding the CC
- [x] Rendering the CC on HTML video element
- [x]  Improve the callback of generating the new cue CC in CTA608 parser in parser cache
- [ ] Move the caption caching and searching logic into SDK
- [ ] Return required caption attributes such as styles and layout

**Note**: we are using `requestVideoFrameCallback` in demo app to render the closed caption, but this API is not available in FireFox so far, so we need to find an alternative solution on firefox.


packages/millicast-sdk/src/workers/TransformWorker.worker.js
```diff
@@ -10,15 +10,17 @@ const synchronizationSources = {}
 let synchronizationSourcesWithMetadata = []
 
 function createReceiverTransform (mid) {
+  const closedCaptionReady = (startTime, endTime, text) => {
+    self.postMessage({ mid, event: 'closedCaption', startTime, endTime, text })
+  }
+  resetCodecs()
```
turbidwater, line 16: question: (non-blocking) @vincentsong Is this resetting the codecs after each group of CC is received, or is this running less often?

packages/millicast-viewer-demo/src/viewer.js
```diff
@@ -242,27 +247,63 @@ document.addEventListener("DOMContentLoaded", () => {
...
   }, 2000);
   subscribe();
 });
 
+function registerRequestVideoFrameCallback(videoElem) {
+  const captionElem = document.getElementById('caption');
+  let curCC = null
+  if ("requestVideoFrameCallback" in HTMLVideoElement.prototype) {
+    const processor = (now, { rtpTimestamp }) => {
+      const ts = rtpTimestamp / 1000;
```
turbidwater, line 263: question: @vincentsong would this work with any Millicast RTS stream or does it need to have timestamp metadata enabled? When I was looking at using this in the hybrid player, I was not getting the accurate timestamps I was hoping for. I was looking for the `captureTime` property, but it was not present in the Millicast RTS stream's frame data. 

Also, does this work in Firefox? I recall it not having full support for the video frame callback.

PR #393, opened 2024-07-08
Change error handling to prevent flow to be cut off

ngietka:
See DIOS-6050 for more info:
https://jira.dolby.net/jira/browse/DIOS-6050

This minor change fixes the bug we've been seeing in which SPS errors during metadata handling cause playback to fail.
We discovered this while working on the Bally's POC when trying to play ads triggered by RTS metadata. Currently the POC will crash on every 3rd refresh or so.


Santiago-Souto (approved):
LGTM

PR #405, opened 2024-07-22
First pass through of using nx

dubeyShivank:
This is the first pass - 

Next stage is to add nx based commands instead of `lerna` based commands.


Santiago-Souto (approved):
LGTM

PR #406, opened 2024-07-24
Add connection duration to diagnostics response

dubeyShivank:


gregdolby (approved):

packages/millicast-sdk/src/types/index.d.ts
```diff
@@ -125,7 +125,14 @@ declare module '@millicast/sdk' {
...
      */
-    stats : Array<Object>
+    stats : Array<Object>,
+
+    /**
+     * Represents how long the stream has been connected for. If a stream disconnects and 
+     * re-connects a few times, this will reflect the duration only since the last time a 
+     * successful connection was established
+     */
+    connectionTime: Number
```
Santiago-Souto, line 135: I know this PR is already closed, but this attribute is not the same as the value returned, in Diagnostics method the attribute returned is `connectionDurationMs`

dubeyShivank: I'll fix it in a follow up PR :) Thanks for picking this up @Santiago-Souto

dubeyShivank: Here you go https://github.com/millicast/millicast-sdk/pull/412

PR #415, opened 2024-08-02
Convert bitrate to bps

Santiago-Souto:
Webrtc stats library has `bitrate` attribute in Bytes per second while the SDK delivered it in bits per second, so we only need to multiply those incoming values times 8 to convert them into bps.

### DIFF
--- a/.changeset/loud-drinks-complain.md
+++ b/.changeset/loud-drinks-complain.md
@@ -2,4 +2,4 @@
 "@millicast/sdk": patch
 ---
 
-Fix bitrate unit. Now showing correctly bits per second.
+Added bitrateBitsPerSecond stats attribute. Now bitrate attribute is shown in Bytes per second and bitrateBitsPerSecond in bits per second.
--- a/.changeset/loud-drinks-complain.md
+++ b/.changeset/loud-drinks-complain.md
@@ -1,0 +1,5 @@
+---
+"@millicast/sdk": patch
+---
+
+Fix bitrate unit. Now showing correctly bits per second.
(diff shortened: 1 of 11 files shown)


FabienLavocat (approved):
Looks OK to me.
Do you think creating a new property like `bitrateBitsPerSecond` would have been better since the `bitrate` property now has the description of `Bytes per second` but will actually be `Bits per second`.

FabienLavocat:
> in the SDK we have the bitrate attribute as bits per second, that's why I just changed that attribute. Does that make sense?

I understand why you did it like that, I'm just thinking that because you override the value of the `bitrate` property, if someone tries to read that value outside of the SDK, then they will have a hard time figuring out that it's not `Bps` anymore but `bps`. So, by adding a new property (e.g. `bitrateBitsPerSecond`) and of course updating the player to display this new property, then the developers will have access to all the data and pick the one they want.

Santiago-Souto:
> So, by adding a new property (e.g. bitrateBitsPerSecond) and of course updating the player to display this new property, then the developers will have access to all the data and pick the one they want.

Seems reasonable, @dubeyShivank what are your thoughts on this?

dubeyShivank:
> > So, by adding a new property (e.g. bitrateBitsPerSecond) and of course updating the player to display this new property, then the developers will have access to all the data and pick the one they want.
> 
> Seems reasonable, @dubeyShivank what are your thoughts on this?

Lets roll with this and I'll close my PR

FabienLavocat (approved):

PR #417, opened 2024-08-08
Reinstate onmetadata

dubeyShivank:


.changeset/eleven-shoes-fix.md
```diff
@@ -0,0 +1,5 @@
+---
+"@millicast/sdk": patch
+---
+
+Bring back onMetadata
```
Santiago-Souto, line 5: This should not be a changesets, as it is a rollback

Santiago-Souto (changes requested):

.changeset/eleven-shoes-fix.md
dubeyShivank: Ok, I shall remove it.

Santiago-Souto (approved):
LGTM

PR #423, opened 2024-08-13
Optional metadata UUID

dubeyShivank:
If there is no UUID in the metadata (as would be the case with SEI PIC timing, do not try to parse it.


FabienLavocat (approved):

Santiago-Souto (approved):
LGTM

PR #429, opened 2024-08-27
Nx workspace typescript

gregdolby:


gregdolby:
All remaining JS files can be converted. PeerConnection.ts and Signaling.ts need to be converted to TS. I switched them to Typescript to help finish View.ts conversion. Use index.d.ts file for types

Santiago-Souto (approved):
LGTM

PR #430, opened 2024-08-28
Remove deprecated property

dubeyShivank:
`StreamName` is removed from everywhere


packages/millicast-sdk/src/utils/BaseWebRTC.js
```diff
@@ -2,6 +2,7 @@ import EventEmitter from 'events'
 import PeerConnection, { webRTCEvents } from '../PeerConnection'
 import { signalingEvents } from '../Signaling'
 import Diagnostics from './Diagnostics'
+import { log } from 'console'
```
Santiago-Souto, line 5: This was a typo, wasn't it?
```suggestion
```

Santiago-Souto (changes requested):

packages/millicast-sdk/src/utils/BaseWebRTC.js
dubeyShivank: Hahh, yes :D

Santiago-Souto (approved):

PR #437, opened 2024-09-12
Refactored webrtc and signaling events management

vincentsong:
To handle a new use case: player is launched while stream with subscribe token is still not live, subscribe endpoint will not return error but 200 OK instead, this PR is to ensure main stream's 'track' event can only be triggered after 'active' event 

Note: the `mediaElement` argument of View's constructor is removed due to conflict with DRM stream workflow


packages/millicast-sdk/src/View.js
```diff
@@ -38,11 +38,10 @@ const connectOptions = {
  * @constructor
  * @param {String} streamName - Deprecated: Millicast existing stream name.
  * @param {tokenGeneratorCallback} tokenGenerator - Callback function executed when a new token is needed.
- * @param {HTMLMediaElement} [mediaElement=null] - Target HTML media element to mount stream.
  * @param {Boolean} [autoReconnect=true] - Enable auto reconnect to stream.
  */
 export default class View extends BaseWebRTC {
-  constructor (streamName, tokenGenerator, mediaElement = null, autoReconnect = true) {
+  constructor (streamName, tokenGenerator, autoReconnect = true) {
```
dubeyShivank, line 44: Like `streamName`, can we mark `mediaElement` as deprecated and handle the same way? Otherwise its a breaking change for players if the constructor of the view class all of a sudden doesn't receive one property.

And handle it the same way as we handle streamName

packages/millicast-sdk/src/View.js
```diff
@@ -452,7 +458,7 @@ export default class View extends BaseWebRTC {
     const drmOptions = {
       merchant: 'dolby',
       // TODO: change to Product when backend is ready
```
dubeyShivank, line 460: TODO can be removed.

packages/millicast-sdk/src/View.js
vincentsong: done

packages/millicast-sdk/src/View.js
vincentsong: removed

packages/millicast-sdk/src/View.js
```diff
@@ -53,11 +53,9 @@ export default class View extends BaseWebRTC {
...
     // mapping media ID of RTCRtcTransceiver to DRM Options
     this.drmOptionsMap = null
-    if (mediaElement) {
-      this.on(webRTCEvents.track, e => {
-        mediaElement.srcObject = e.streams[0]
-      })
-    }
+    // cache of events to coordinate re-emitting
+    this.eventQueue = []
+    this.isMainStreamActive = false
```
dubeyShivank, line 58: Sorry couldn't figure out how to do this on GitHub around line 49, so sharing here. 
```suggestion
    this.isMainStreamActive = false;
        if (mediaElement) {
      logger.warn('The mediaElement property has been deprecated. In a future release, this will be removed. Please do not rely on this value. Instead, do this in either the `track` or the `active` broadcast event.')
    }
```

vincentsong: Done

dubeyShivank (approved):

PR #442, opened 2024-09-24
try a diff config to see if nx works

dubeyShivank:


package.json
```diff
@@ -49,10 +47,11 @@
...
     "husky": "^4.3.8",
     "install": "^0.13.0",
-    "lerna": "^5.5.0",
+    "lerna": "^8.1.8",
     "lint-staged": "^10.5.4",
     "netlify-cli": "^17.15.6",
     "npm": "^8.11.0",
-    "vite": "^5.2.10"
+    "vite": "^5.2.10",
+    "nx": "16.10.0"
```
vincentsong, line 55: we should use nx version >=17.1.2 if we are using Lerna 8.1.8
https://lerna.js.org/docs/lerna-and-nx-version-matrix

dubeyShivank: Thanks! Will upgrade :)

PR #446, opened 2024-09-25
Hotfix: Exposed DRM option target latency

vincentsong:


packages/millicast-sdk/src/types/index.d.ts
```diff
@@ -1424,6 +1424,12 @@ declare module '@millicast/sdk' {
 
     /** The audio media ID of RTCRtpTransceiver */
     audioMid?: string;
+
+    /** The average target latency, Can be set to 0,
```
dubeyShivank, line 1428:
```suggestion
    /** The average target latency. It can be set to 0, 
```

packages/millicast-sdk/src/types/index.d.ts
```diff
@@ -1424,6 +1424,12 @@ declare module '@millicast/sdk' {
 
     /** The audio media ID of RTCRtpTransceiver */
     audioMid?: string;
+
+    /** The average target latency, Can be set to 0, 
+     * enabling zero-buffering mode (which is not recommended as it affects video playback smoothness). 
+     * The default value is 100 ms, except when PlayReady or Widevine L1 on Windows are used - those
```
dubeyShivank, line 1430: non blocking, but because we don't support PlayReady, we can remove it from here

dubeyShivank (approved):

packages/millicast-sdk/src/View.js
```diff
@@ -465,6 +466,7 @@ export default class View extends BaseWebRTC {
       customTransform: this.options.metadata,
       videoElement: options.videoElement,
       audioElement: options.audioElement,
+      mediaBufferMs: options.mediaBufferMs || 100,
```
mildsunrise, line 469: I suggest not supplying a default ourselves, and leave it to their SDK to set the fallback value... mainly because they could (now or in the future) be using smarter heuristics to set the default

```suggestion
      mediaBufferMs: options.mediaBufferMs,
```

gregdolby (dismissed):

packages/millicast-sdk/src/types/index.d.ts
vincentsong: I just copied this from castlab SDK's integration guid, and there is also `play ready` url from director subscribe API's response

packages/millicast-sdk/src/View.js
vincentsong: OK, I will conditionally set this options instead of setting 100ms

packages/millicast-sdk/src/types/index.d.ts
vincentsong: Done.

packages/millicast-sdk/src/View.js
vincentsong: Done

dubeyShivank (approved):

PR #447, opened 2024-09-26
Fix for subscriber token not set in getSubscriber's response

adamallaf:
Subscriber token is not set in getSubscriber's response when getSubscriber args are passed as an object.


Santiago-Souto (approved):
LGTM

turbidwater (approved):

PR #449, opened 2024-09-30
Fix build tests

gregdolby:
This PR fixes npm run build/test/start. Moving forward let us make PR's into the nx-workspace-typescript branch from our own branch. And before we merge the PR validate that build, test, start work and that we have successful playback in the viewer demo app 'http://localhost:10002/?accountId=k9Mwad&streamName=multiview'

I also switched linting to only target millicast sdk package and not all demo apps for now.

Steps after this PR is merged:
1. Fix all linting errors
2. Validate that index.d.ts file generated from ts build matches the manual type file from v0


packages/millicast-sdk/src/Signaling.ts
```diff
@@ -229,29 +241,29 @@ export default class Signaling extends EventEmitter {
       if (optionsParsed.disableVideo && optionsParsed.disableAudio) {
         throw new Error('Not attempting to connect as video and audio are disabled')
       }
+      await this.connect()
```
gregdolby, line 244: This was breaking tests all offering publishing stream tests and it took me a whole day to find lol. The TransactionManager is null until we try and connect. So we have to connect before the null check.

Santiago-Souto (approved):
LGTM

PR #450, opened 2024-10-03
Fix lint

gregdolby:
This is a a fully converted TS SDK which builds/tests/starts and lints.

@Santiago-Souto could you take a good look over TransformWorker.worker.ts and Codecs.ts. I tried my best at typing those files using WebRTC type from MDN and reverse engineering the data in these files but I'm still not 100%.


packages/millicast-sdk/src/View.ts
```diff
@@ -467,13 +479,14 @@ export default class View extends BaseWebRTC {
...
 
-  async onRtcDrmFetch(url: string, opts: any) {
-    if (!opts.headers) {
-      opts.headers = new Headers()
-    }
+  async onRtcDrmFetch(url: string, opts: RequestInit) {
+    opts.headers = (opts.headers as Headers) || new Headers()
+    // if (!opts.headers) {
+    //   opts.headers = new Headers()
+    // }
```
Santiago-Souto, line 489: Is this meant to be left commented out?

packages/millicast-sdk/src/utils/Codecs.ts
```diff
@@ -64,28 +38,43 @@ export const DOLBY_SEI_TIMESTAMP_UUID = '9a21f3be-31f0-4b78-b0be-c7f7dbb97250'
 // When the SDK inserts its own timecode into the unregistered block, it uses this identifier
 export const DOLBY_SDK_TIMESTAMP_UUID = 'd40e38ea-d419-4c62-94ed-20ac37b4e4fa'
 
+export interface SeiMetadata {
+  seiPicTimingTimeCodeArray: SeiPicTimingTimeCode[]
```
Santiago-Souto, line 42: Shouldn't this be optional? As it is being set in a switch case

packages/millicast-sdk/src/View.ts
```diff
@@ -466,13 +479,14 @@ export default class View extends BaseWebRTC {
...
```
Santiago-Souto, line 489: Is this intended to be commented out?

Santiago-Souto (changes requested):

packages/millicast-sdk/src/View.ts
gregdolby: Nope, good catch

packages/millicast-sdk/src/utils/Codecs.ts
gregdolby: I canged this to optional.

Santiago-Souto (approved):
Sample apps are not working but probably due to being in TS not fully converted

gregdolby:
> Sample apps are not working but probably due to being in TS not fully converted

Sample apps should be working. Try runing 'nx reset' delete node_modules and package-lock. Also be sure to delete all node_modules in pacakges. The only node_modules that exist any more are in project root. I just validated that the apps work

PR #451, opened 2024-10-10
Update latest castlabs SDK that better works on react-native

dubeyShivank:


Santiago-Souto (approved):
If this new SDK is backwards compatible, go ahead! I would want @gregdolby or @vincentsong to validate as well

dubeyShivank:
@gregdolby @vincentsong I've done an `npm run build` and an `npm run test` and they seem to work. Can you please publish an encrypted stream and we can test out if there are no breaking changes in this one?

gregdolby (approved):
This looks good to me. I'm just not sure if there is anything that was added to RC-12 that wouldn't be in this RC-10 version

gregdolby (approved):

gregdolby (approved):

PR #453, opened 2024-10-15
refactor: reduce bundle size

gregdolby:
- Removed 'joi'
- created custom publish options validation (our joi validation wasn't doing much in the first place, just warnings. all undefined properties would be validated)
- added tests for custom validation
- removed excess bundled files. only es module build with types exists now


dubeyShivank (approved):

gregdolby:
I am working through this e2e test failure and it has to do with the build output. Puppeteer needs umd build but I am trying to figure out how to get it to work with es module bundle.

gregdolby:
I ended up adding umd bundle back. Jest tests and puppeteer run in node and they both support esm but for jest it is experimental.

PR #455, opened 2024-10-16
chore: migrate from changesets to nx release

Santiago-Souto:
NX Release will use conventional commits as a guide to create changelogs and version numbers. First runs will be manual, but the idea is to automate them in the future. Let's discuss on the contributing file and if it needs some more information


CONTRIBUTING.md
```diff
@@ -17,9 +17,40 @@ Asumming that you have Node 12.10.x or newer and `npm` installed, install the re
 npm ci
 ```
 
-### Changesets
+### Pull request - Conventional Commits
```
dubeyShivank, line 20: is there a pre-commit hook we can leverage to achieve this?

dubeyShivank: Something like [this](https://dev.to/mbarzeev/a-git-hook-for-commit-messages-validation-no-husky-just-js-1hni) or [this](https://github.com/compilerla/conventional-pre-commit)

Santiago-Souto: I don't feel getting every commit into conventional commit standard would be beneficial, as for each change to main a PR is created, and commits are squashed, leaving the commit message in main as the PR title, therefore I feel checking the title would be best. There are some actions in Github for that kind of stuff. I'll add it so that we can check if it is worth the use

Santiago-Souto: @dubeyShivank do you feel the GH Action I added would be enough?

dubeyShivank: let me check it out :)

.github/actions/PRConventionalCommits-1.3.0/action.yaml
```diff
@@ -0,0 +1,27 @@
+name: 'Conventional Commit In Pull Requests'
+description: 'Check that the PR title contains a valid task type and (optionally) a task number, and label the PR based on the task type'
+inputs:
+
+  task_types:
+    description: 'An array of task types. Example: ["feat","fix","docs","test","ci","refactor","perf","chore","revert"]'
+    required: true
+  ticket_key_regex:
+    description: 'Regular expression to match issue number in PR title, by default does not validate. Example ^PROJECT-\d{2,5}$'
```
dubeyShivank, line 9: Do we need this input? Because we don't generally have a ticket pattern

dubeyShivank: Unless its just the standard thing that comes with the lib...

Santiago-Souto: That's something from the action's repository. There's an option to check for project/ticket pattern but for the config we put, we won't be using that option

CONTRIBUTING.md
dubeyShivank: Now that I've seen the template, yeah should be good

dubeyShivank (approved):

PR #456, opened 2024-10-17
Call subscribe again when there is no cached DRM profile

vincentsong:
We need to reproduce the issue when open viewer before broadcast start, the first 'subscrbe' API returned 200 response without DRM profile, when 'active' event comes we need to invoke 'subscdribe' API again.

But I cannot reproduce that case, since the 'subscribe' API always return 400 now.

### DIFF
--- a/packages/millicast-sdk/src/View.js
+++ b/packages/millicast-sdk/src/View.js
@@ -246,10 +246,6 @@ export default class View extends BaseWebRTC {
       streamName: this.streamName,
       url: `${subscriberData.urls[0]}?token=${subscriberData.jwt}`
     })
-    if (subscriberData.drmObject) {
-      // cache the DRM license server URLs
-      this.DRMProfile = subscriberData.drmObject
-    }
     if (subscriberData.subscriberToken) {
       this.subscriberToken = subscriberData.subscriberToken
     }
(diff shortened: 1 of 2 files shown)


vincentsong:
> We need to reproduce the issue when open viewer before broadcast start, the first 'subscrbe' API returned 200 response without DRM profile, when 'active' event comes we need to invoke 'subscdribe' API again.
> 
> But I cannot reproduce that case, since the 'subscribe' API always return 400 now.

We changed the logic, only use the drmObject from 2nd subscribe endpoint's response. So we could verify it is working

Note: this is a patch, we should use the `drmObject` from `active` event in next release

packages/millicast-sdk/src/View.js
```diff
@@ -306,21 +300,32 @@ export default class View extends BaseWebRTC {
...
+              if (subscriberData.drmObject) {
+                // cache the DRM license server URLs
+                this.DRMProfile = subscriberData.drmObject
+              }
+            }
+            this.emit(signalingEvents.broadcastEvent, event)
             this.isMainStreamActive = true
-            while (this.eventQueue.length > 0) {
-              this.onTrackEvent(this.eventQueue.shift())
+            if (event.name === 'active') {
```
dubeyShivank, line 316: Question - this is the `active` event, so why do we need the `if` statement?

dubeyShivank (approved):

PR #460, opened 2024-10-21
fix: regression on broadcast events

Santiago-Souto:
Add removed reemition of broadcast events as it was causing trouble with events being emitted

### DIFF
--- a/packages/millicast-sdk/src/View.js
+++ b/packages/millicast-sdk/src/View.js
@@ -311,11 +311,12 @@ export default class View extends BaseWebRTC {
                 this.DRMProfile = subscriberData.drmObject
               }
             }
+            this.emit(signalingEvents.broadcastEvent, event)
             this.isMainStreamActive = true
             while (this.eventQueue.length > 0) {
               this.onTrackEvent(this.eventQueue.shift())
             }
-            break
+            return
           case 'inactive':
             this.isMainStreamActive = false
             break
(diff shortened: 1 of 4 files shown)


vincentsong (dismissed):
I recall I already removed these, no idea why they are still in the code

vincentsong (commented):
Sorry, we cannot add this back, otherwise, user will get active event twice

packages/millicast-sdk/src/View.js
```diff
@@ -254,8 +254,10 @@ export default class View extends BaseWebRTC {
     await webRTCPeerInstance.createRTCPeer(this.options.peerConfig)
     // Stop emiting events from the previous instances
     this.stopReemitingWebRTCPeerInstanceEvents?.()
+    this.stopReemitingSignalingInstanceEvents?.()
     // And start emitting from the new ones
     this.stopReemitingWebRTCPeerInstanceEvents = reemit(webRTCPeerInstance, this, Object.values(webRTCEvents).filter(e => e !== webRTCEvents.track))
+    this.stopReemitingSignalingInstanceEvents = reemit(signalingInstance, this, [signalingEvents.broadcastEvent])
```
vincentsong, line 260: We should not call reemit here since we already hooked these event in 'initConnection' and will reemit these events manually after processing them

vincentsong (changes requested):

packages/millicast-sdk/src/View.js
dubeyShivank: So @vincentsong remove this line, but keep line 257? Is that the suggested change?

Santiago-Souto: @vincentsong The latest change should do the trick. The problem was that where the new listener was emitting the events, it was only emitting them when `event.data.sourceId` is null. I moved the emitter outside the if condition. Let's discuss if the change between the return and break it's OK or the emit in that case should be before the while loop

Santiago-Souto: I think I answered my question, will put the emitter back before the while loop

Santiago-Souto:
Just saw this fix was already done, closing PR

PR #462, opened 2024-10-22
bugfix: added default events into signaling service

vincentsong:


packages/millicast-sdk/src/Signaling.js
```diff
@@ -196,7 +196,9 @@ export default class Signaling extends EventEmitter {
     // Signaling server only recognizes 'AV1' and not 'AV1X'
     sdp = SdpParser.adaptCodecName(sdp, 'AV1X', VideoCodec.AV1)
 
-    const data = { sdp, streamId: this.streamName, pinnedSourceId: optionsParsed.pinnedSourceId, excludedSourceIds: optionsParsed.excludedSourceIds }
+    // default events
+    const events = ['active', 'inative', 'layers', 'viewercount', 'vad', 'updated', 'migrate', 'stopped']
+    const data = { sdp, streamId: this.streamName, pinnedSourceId: optionsParsed.pinnedSourceId, excludedSourceIds: optionsParsed.excludedSourceIds, events }
 
     if (optionsParsed.vad) { data.vad = true }
     if (Array.isArray(optionsParsed.events)) { data.events = optionsParsed.events }
```
dubeyShivank, line 204: Can we do it this way? This will prevent user defaults
```suggestion
    if (Array.isArray(optionsParsed.events)) { 
        data.events = optionsParsed.events 
    } else {
        data.events = events     
    }
```

dubeyShivank, line 204:
```suggestion
    // default events
    const events = ['active', 'inative', 'layers', 'viewercount', 'vad', 'updated', 'migrate', 'stopped']
    const data = { sdp, streamId: this.streamName, pinnedSourceId: optionsParsed.pinnedSourceId, excludedSourceIds: optionsParsed.excludedSourceIds }

    if (optionsParsed.vad) { data.vad = true }
        if (Array.isArray(optionsParsed.events)) { data.events = optionsParsed.events }
    if (Array.isArray(optionsParsed.events)) { 
        data.events = optionsParsed.events 
    } else {
        data.events = events     
    }
```

vincentsong: your code is same as my code, and I prefer mine.

vincentsong: My concern is this patch won't work as expected when media server changed the default events list. So that's why I called it a patch

dubeyShivank: > your code is same as my code, and I prefer mine.
Sorry read it in a rush. You're correct, it's the same either way

dubeyShivank (approved):

PR #464, opened 2024-10-23
fix (millicast-viewer-demo): Remove autoReconnect null set in instance creation

Santiago-Souto:


jwrobdolby:
Looks good to me, tested - OK

jwrobdolby (approved):

PR #465, opened 2024-10-24
fix: no video in multi-source stream when main source id is not null

vincentsong:


packages/millicast-viewer-demo/src/viewer.js
```diff
@@ -81,9 +81,6 @@ const newViewer = () => {
         millicastView.configureDRM(drmOptions);
       }
     }
-    let layers = event.data["layers"] !== null ? event.data["layers"] : {};
-    if (event.name === "layers" && Object.keys(layers).length <= 0) {
-    }
```
dubeyShivank, line 86: Was this for testing purposes?

dubeyShivank (approved):

packages/millicast-viewer-demo/src/viewer.js
vincentsong: It's useless now, so remove it from demo

dubeyShivank (approved):

PR #466, opened 2024-10-24
test: integration tests for PR pipeline

jwrobdolby:
Initial integration tests of WebSDK to be run in PR pipeline. Next steps are:
- [x] create token before each run to avoid streamName conflicts
- [ ] provide env variables used by tests
- [ ] extend github workflow to run on rp2 and staging


packages/millicast-sdk/integration-tests/tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
```
dubeyShivank, line 1: Question - does this need its own tsconfig? Won't it just use the one used by `packages/millicast-sdk` ?

@gregdolby @Santiago-Souto @vishalkharge if you guys know?

packages/millicast-sdk/integration-tests/src/support-utils/utils.ts
```diff
@@ -0,0 +1,15 @@
+export function parseData(data: Record<string, any>) {
+  const optionsDict: Record<string, any> = {};
+  //convert strings into boolean if true/false encountered
+  Object.entries(data).forEach(([key, value]) => {
+    if (value === "true" || value === "false") {
```
dubeyShivank, line 5: should this be 
```suggestion
    if (value.toLowerCase() === "true" || value.toLowerCase() === "false") {
```

dubeyShivank: depends, though, I'm not quite sure

packages/millicast-sdk/integration-tests/src/support-utils/generic.ts
```diff
@@ -0,0 +1,31 @@
...
+  
+    while (new Date().getTime() - startTime < timeout) {
+      try {
+        result = await func()
+        if (result === false) return result
+      } catch {}
+      await sleep(500)
+    }
+    return result
+  }
```
dubeyShivank, line 31:
```suggestion
export async function retryUntilTrue(func: () => Promise<boolean>, timeoutSeconds = 10): Promise<boolean> {
  return retryUntil(func, timeout, true);

  }
  
  export async function retryUntilFalse(func: () => Promise<boolean>, timeoutSeconds = 10): Promise<boolean> {
  return retryUntil(func, timeout, false);
  }
  
  async function retryUntil(func: () => Promise<boolean>, timeoutSeconds = 10, condition: boolean = false): Promise<boolean> {
    const timeout = timeoutSeconds * 1000
    const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms))
    const startTime = new Date().getTime()
    let result = true

    while (new Date().getTime() - startTime < timeout) {
      try {
        result = await func()
        if (result) return result
      } catch {}
      await sleep(500)
    }
    return result
  }
  
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,78 @@
+name: Integration Tests
```
dubeyShivank, line 1: Question(s) maybe for @vishalkharge 
- This isn't running on a docker image, is it? It's running on GH runners, so should we be mindful of the number of tests and the number of re-tries we perform?

packages/millicast-sdk/integration-tests/features/connect.feature
```diff
@@ -0,0 +1,125 @@
...
+    When the "publisher1" stops the published stream
+    Then the "publisher1" stream should be NOT LIVE
+
+    When the "publisher1" starts the stream with the specified options
+      | codec | h264 |
+    Then the "publisher1" stream should be LIVE
+    And the "viewer1" connected stream should be LIVE
+    And the "viewer1" should be able to view media tracks for the connected stream
+
+  Scenario: Publisher connects with disableVideo as <disableVideo> and disableAudio as <disableAudio> - Expected Fail DIOS-7279
```
dubeyShivank, line 83: No DIOS-* references please, open source repo :)

packages/millicast-sdk/integration-tests/src/support-utils/utils.ts
jwrobdolby: That's right, more flexible. Done

packages/millicast-sdk/integration-tests/src/support-utils/generic.ts
jwrobdolby: Nice, works ok but with adjustments:
- I have to put timeoutSeconds instead of timeout when calling retryUntil()
- 'if (condition === result) return result' in while loop
without it test failed or passed after first iteration. Please have a look @dubeyShivank 

```suggestion
export async function retryUntilTrue(func: () => Promise<boolean>, timeoutSeconds = 10): Promise<boolean> {
  return retryUntil(func, timeoutSeconds, true);
  }
  
export async function retryUntilFalse(func: () => Promise<boolean>, timeoutSeconds = 10): Promise<boolean> {
  return retryUntil(func, timeoutSeconds, false);
  }
  
  async function retryUntil(func: () => Promise<boolean>, timeoutSeconds = 10, condition = false): Promise<boolean> {
    const timeout = timeoutSeconds * 1000
    const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms))
    const startTime = new Date().getTime()
    let result = true
    while (new Date().getTime() - startTime < timeout) {
      try {
        result = await func()
        if (condition === result) return result
      } catch {}
      await sleep(500)
    }
    return result
  }
```

packages/millicast-sdk/integration-tests/tsconfig.json
gregdolby: It looks like it is reccomended for playwright. lhttps://playwright.dev/docs/test-typescript#tsconfigjson

gregdolby:
Hey @jwrobdolby this PR looks great! Do you know what I would need to do to have it run locally? Also, it might be worth adding the files generated when the test runs to the .gitignore

packages/millicast-sdk/integration-tests/src/support-utils/utils.ts
dubeyShivank: Probably not pushed :)

packages/millicast-sdk/integration-tests/features/simulcast.feature
```diff
@@ -0,0 +1,28 @@
...
+        Then the "viewer1" connected stream should be LIVE
+
+        When the "viewer1" selects simulcast layer with encodingId "0"
+        Then the "viewer1" verifies video resolution for layer "0"
+
+        When the "viewer1" selects simulcast layer with encodingId "1"
+        Then the "viewer1" verifies video resolution for layer "1"
+
+        When the "viewer1" selects simulcast layer with encodingId "2"
+        Then the "viewer1" verifies video resolution for layer "2"
```
Santiago-Souto, line 23: Instead of 0,1,2 this should be f,h,q

Santiago-Souto (changes requested):

PR #467, opened 2024-10-25
Remove setSimulcast function

juanpiquerez:
Simulcast is currently supported on Chrome only, change this behaviour so that it works on all browsers


dubeyShivank (approved):

Santiago-Souto (approved):
LGTM

PR #475, opened 2024-10-31
Fix Inconsistencies with disableVideo/disableAudio

juanpiquerez:
Fix when connect publisher with ({disableVideo:true}), viewer does not play audio.


Santiago-Souto (approved):
LGTM

dubeyShivank (approved):

PR #490, opened 2025-06-12
Add force Smoothing to Viewer connect options

dubeyShivank:


FabienLavocat (approved):

packages/millicast-sdk/src/View.js
```diff
@@ -94,6 +95,7 @@ export default class View extends BaseWebRTC {
    * @param {Number} [options.forcePlayoutDelay.min]    - Set minimum playout delay value.
    * @param {Number} [options.forcePlayoutDelay.max]    - Set maximum playout delay value.
    * @param {Boolean} [options.enableDRM]               - Enable DRM, default is false.
+   * @param {Boolean} [options.forceSmooth]             - Enables/Disables force smoothing (less aggressive layer switching) when viewing streams.
```
FabienLavocat, line 98: Add what is the default value.

FabienLavocat (changes requested):

packages/millicast-sdk/src/View.js
```diff
@@ -23,7 +23,8 @@ const connectOptions = {
   peerConfig: {
     autoInitStats: true,
     statsIntervalMs: 1000
-  }
+  },
+  forceSmooth: true
```
bcostdolby, line 27: Is this the default value? We should make it false by default, or better yet omitted and let the server choose the default and only include on the URL if it is specified by the user?

packages/millicast-sdk/src/View.js
```diff
@@ -83,6 +84,26 @@ export default class View extends BaseWebRTC {
    * Connects to an active stream as subscriber.
    *
    * In the example, `addStreamToYourVideoTag` and `getYourSubscriberConnectionPath` is your own implementation.
+<<<<<<< feature/force-smooth
```
bcostdolby, line 87: Bad merge here I think.

packages/millicast-sdk/src/types/index.d.ts
```diff
@@ -1191,6 +1191,21 @@ declare module '@millicast/sdk' {
...
       max?: number;
     };
+
+    /**
+     * Forces video to be sent on downlink when switching from higher quality
+     * layers.
+     *
+     * If true, smooth transitions will be enabled when changing from a higher
+     * rate to a lower rate, this may cause congested use cases to take longer to
+     * recover.
```
bcostdolby, line 1201: this may cause congested use cases to take longer to recover at the benefit of minor glitches in the network not causing pauses in the video content.

PR #496, opened 2025-07-16
Support frame metadata extraction on older browsers.

dubeyShivank:


packages/millicast-sdk/src/workers/TransformWorker.worker.js
```diff
@@ -36,6 +35,33 @@ function createReceiverTransform (mid) {
...
+    } catch (e) {
+      // instanceof might fail in some contexts
+    }
+  }
+
+  if (frame.constructor && frame.constructor.name === 'RTCEncodedVideoFrame') {
+    return true
+  }
+
+  if (typeof frame.width === 'number' && typeof frame.height === 'number') {
```
GillesMoris-Dolby, line 54: Should we explicitly check all properties and methods that we need? Such as `getMetadata()` as well

PR #524, opened 2026-08-24
feat(sdk): add maintainResolution publish option to hold a constant encoder resolution

bcostdolby:
## Summary

Adds an opt-in `maintainResolution` publish option (default `false`) that asks the pipeline to hold the encoder at a constant resolution and give up frame rate instead.

```js
await publisher.connect({ mediaStream, maintainResolution: true })
```

**Why.** This is a workaround for decoders that glitch on a mid-stream resolution change — we hit it relaying a browser publisher to a third-party RTMP destination (YouTube), where the publisher ramps 1280x720 → 960x540 → 1280x720 within the first ~7 seconds while the bandwidth estimate settles, and the downstream output is corrupted for minutes afterwards.

The mechanism: a resolution change forces the H.264 encoder to re-emit its parameter sets, and `max_num_ref_frames` changes along with the resolution because it is bounded by the level's DPB capacity divided by the frame size. Some decoders mishandle a stream where that value changes mid-flight. `max_num_ref_frames` is not settable from the browser — there is no API for it — so the only lever we have is to remove the trigger, i.e. keep the resolution constant so the parameter sets never need to be re-emitted in the first place. (Note this is not specific to any one encoder; VideoToolbox is simply where we observed it. Nor can the SDK select the encoder implementation — no browser exposes that to JS.)

Until now publishers had no way to ask for a stable resolution: neither `contentHint` nor `degradationPreference` was reachable through the SDK.

**Why both properties, not one.** Neither is reliable alone. `contentHint = 'detail'` is broadly supported and steers libwebrtc's degradation preference implicitly; `degradationPreference = 'maintain-resolution'` states it explicitly but is not implemented everywhere. Both are set, both are read back, and a `logger.warn` fires if a browser silently ignored one instead of leaving the caller to assume it worked. A rejected `setParameters()` is caught and does not fail the broadcast — the option degrades to `contentHint` alone.

`Publish.initConnection()`:

```js
// after the "MediaStream required" validation and the duplicate-connect isActive() guard,
// so a rejected connect() never mutates a live track's hint
if (this.options.maintainResolution) applyResolutionContentHint(this.options.mediaStream)
...
// after getRTCLocalSDP() resolves — that is what adds the tracks, so senders do not exist before it
if (this.options.maintainResolution) await applyDegradationPreference(webRTCPeerInstance.getRTCPeer())
```

The sender pass uses the local `webRTCPeerInstance` rather than `this.getRTCPeerConnection()` so it is also correct on the `data.migrate` path, where `this.webRTCPeer` still points at the previous peer. `applyResolutionContentHint` handles `mediaStream` being either a `MediaStream` or an array of tracks, matching what the rest of the option surface accepts.

**Why `replaceTrack` had to change.** `contentHint` is a property of the track, not of the sender, so replacing a track (e.g. a camera switch) drops it and the option would silently stop applying mid-broadcast. `PeerConnection.replaceTrack()` now carries the outgoing track's hint over to the incoming one unless the incoming track declares its own. This is unconditional rather than gated on the new option: preserving an explicitly set hint is correct regardless, and it is inert when no hint was set. `degradationPreference` needs no equivalent — it lives on the sender and survives replacement.

**Caveat.** Both settings are preferences, not guarantees; a browser may still adapt under sustained pressure. On a constrained uplink this trades frame rate for resolution, so a publisher that would previously have shrunk the picture will instead drop frames.

Also: `maintainResolution?: boolean` added to `PublishConnectOptions` in `src/types/index.d.ts`, plus a `minor` changeset. The user-facing docs describe the option by intent (a decoder-glitch workaround) and deliberately leave `max_num_ref_frames` out — that detail lives here rather than in the source.

## Tests

New `MaintainResolution` feature (jest-cucumber): option enabled sets both properties; option absent leaves both untouched; a browser whose `setParameters()` rejects still connects and still gets the content hint. `ChangeMediaTrack` gains two scenarios covering hint carry-over and non-override of an explicit incoming hint. The shared `MockRTCPeerConnection` senders gained `getParameters`/`setParameters` (on both the `addTrack` and `addTransceiver` sender objects) so the sender path is testable.

Verified: `npx jest --testMatch "**/unit/*.steps.js"` → 30 suites / 178 tests pass; `npx eslint src/Publish.js` clean; `npx tsc --build` clean.

Note: the `e2e-test` CI job fails on this branch with `Unauthorized: Account disabled`, which also fails identically on `main` — a test-account issue, not a code issue.

Link to Devin session: https://dolby.devinenterprise.com/sessions/9709b96479024ecb90873c68dadf5535
Requested by: @bcostdolby

---

<a href="https://dolby.devinenterprise.com/review/millicast/millicast-sdk/pull/524" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/olive-donkeys-shave.md
+++ b/.changeset/olive-donkeys-shave.md
@@ -1,0 +1,5 @@
+---
+"@millicast/sdk": minor
+---
+
+Added a `maintainResolution` publish option that keeps the encoder at a constant resolution, giving up frame rate instead when bandwidth or CPU is constrained. `contentHint` is now also carried over to the replacement track by `replaceTrack`.
--- a/.changeset/olive-donkeys-shave.md
+++ b/.changeset/olive-donkeys-shave.md
@@ -2,4 +2,4 @@
 "@millicast/sdk": minor
 ---
 
-Added a `maintainResolution` publish option that keeps the encoder at a constant resolution, giving up frame rate instead when bandwidth or CPU is constrained. `contentHint` is now also carried over to the replacement track by `replaceTrack`.
+Added a `maintainResolution` publish option that works around decoders which glitch on a mid-stream resolution change, by keeping the encoder at a constant resolution and giving up frame rate instead when bandwidth or CPU is constrained. `contentHint` is now also carried over to the replacement track by `replaceTrack`.
--- a/packages/millicast-sdk/src/Publish.js
+++ b/packages/millicast-sdk/src/Publish.js
@@ -32,10 +32,8 @@ const connectOptions = {
 /**
  * Ask the pipeline to keep resolution constant and sacrifice frame rate instead.
  *
- * A mid-stream resolution change forces the encoder to re-emit its parameter sets, and
- * max_num_ref_frames is derived from the level's DPB capacity divided by the frame size, so it
- * changes with the resolution. Some downstream transcoders mishandle that combination. Holding
- * the resolution steady avoids the situation entirely.
+ * This works around decoders that glitch on a mid-stream resolution change; holding the
+ * resolution steady avoids the change in the first place.
  *
  * Applied in two places because neither alone is reliable across browsers: contentHint is
  * broadly supported and steers libwebrtc's degradation preference implicitly, while
@@ -122,10 +120,10 @@ export default class Publish extends BaseWebRTC {
    * @param {Boolean} [options.simulcast = false] - Enable simulcast. **Only available in Chromium based browsers and with H.264 or VP8 video codecs.**
    * @param {String} [options.scalabilityMode = null] - Selected scalability mode. You can get the available capabilities using <a href="PeerConnection#.getCapabilities">PeerConnection.getCapabilities</a> method.
    * **Only available in Google Chrome.**
-   * @param {Boolean} [options.maintainResolution = false] - Hold the encoder at a constant resolution,
-   * giving up frame rate instead when bandwidth or CPU is constrained. Sets `contentHint = 'detail'` on
-   * the video track and `degradationPreference = 'maintain-resolution'` on the sender. Both are
-   * preferences, not guarantees; the browser may still adapt under sustained pressure.
+   * @param {Boolean} [options.maintainResolution = false] - Work around decoders that glitch on a
+   * mid-stream resolution change by holding the encoder at a constant resolution, giving up frame
+   * rate instead when bandwidth or CPU is constrained. Both settings it applies are preferences,
+   * not guarantees; the browser may still adapt under sustained pressure.
    * @param {PeerConnectionConfig} [options.peerConfig = null] - Options to configure the new RTCPeerConnection.
    * @param {Boolean} [options.record = false ] - Enable stream recording. If record is not provided, use default Token configuration. **Only available in Tokens with recording enabled.**
    * @param {Array<String>} [options.events = null] - Specify which events will be delivered by the server (any of "active" | "inactive" | "viewercount").*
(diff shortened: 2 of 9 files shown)


packages/millicast-sdk/src/Publish.js
```diff
@@ -22,12 +22,62 @@ const connectOptions = {
...
     autoInitStats: true,
     statsIntervalMs: 1000
   }
 }
 
+/**
+ * Ask the pipeline to keep resolution constant and sacrifice frame rate instead.
+ *
+ * A mid-stream resolution change forces the encoder to re-emit its parameter sets, and
+ * max_num_ref_frames is derived from the level's DPB capacity divided by the frame size, so it
```
bcostdolby: Its not clear why max_num_ref_frames is important. I think we can avoid mentioning it except in the PR description. 

In the PR description, mention it but also describe why it is important this stays the same. I.e. Some decoders have glitches in the output if this changes, we cant set it directly but by holding the resolution constant we avoid it changing for the VideoToolbox encoder.

packages/millicast-sdk/src/types/index.d.ts
```diff
@@ -1297,6 +1297,13 @@ declare module "@millicast/sdk" {
          * **Only available in Google Chrome.**
          */
         scalabilityMode?: string
+        /**
+         * - Hold the encoder at a constant resolution, giving up frame rate instead when bandwidth
+         * or CPU is constrained. Sets `contentHint = 'detail'` on the video track and
+         * `degradationPreference = 'maintain-resolution'` on the sender. Both are preferences, not
+         * guarantees; the browser may still adapt under sustained pressure.
```
bcostdolby: Make clear this options is primarily to work-around a bug found in some decoders (such as youtube after the webrtc is restreamed).

packages/millicast-sdk/src/Publish.js
```diff
@@ -167,6 +221,9 @@ export default class Publish extends BaseWebRTC {
       logger.error('Error while broadcasting. MediaStream required')
       throw new Error('MediaStream required')
     }
+    if (this.options.maintainResolution) {
+      applyResolutionContentHint(this.options.mediaStream)
+    }
```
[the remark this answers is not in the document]
bcostdolby: I need more clarity on what is going on here. What is the problem with moving this after the is active guard?

.changeset/olive-donkeys-shave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@millicast/sdk": minor
+---
+
+Added a `maintainResolution` publish option that keeps the encoder at a constant resolution, giving up frame rate instead when bandwidth or CPU is constrained. `contentHint` is now also carried over to the replacement track by `replaceTrack`.
```
bcostdolby: Please answer my question instead of making code changes at the moment.

The whole point of this PR is to help work-around a bug in youtube decoding when decoding resolution changes that come from the video toolbox encoder by avoiding resolution changes in the first place.

Is there instead an option we could implement to change the encoder used in the first place to something different?

I.e. In the SDK allow the customer to change to use a different encoder?

packages/millicast-sdk/src/Publish.js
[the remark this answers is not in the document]
bcostdolby: Please move it

packages/millicast-sdk/src/Publish.js
bcostdolby: Please update

packages/millicast-sdk/src/types/index.d.ts
bcostdolby: Please make sure this update is applied

